PR #26, opened 2021-03-09
update helloworld example for Reunion 0.5

huichen123:


BenJKuhn:
This looks great. Can you hold this open, and update the package names one more time when we have the final "release-ready" versions, which will drop the date & CI from the package names? Then update once those are published & I'll sign off.

Scottj1s:
We should update this and DWriteCore sample to look more like the VSIX project templates - remove PS1, wap customizations, etc

Scottj1s (approved):

PR #33, opened 2021-04-29
Fix build configuration for C# app and target framework version of the lib.

rohanp-msft:
If you try to build the C# app as is after cloning the repo, the wapproj won't be built. Deploy won't work either. This is because these things weren't enabled for Any CPU, which is the configuration when you open the solution.

Also, this change updates the target framework version of the lib to match the consuming app. An app dev shouldn't have to get an older version of the SDK unnecessarily.

**How verified**
Built the lib.
Built and ran the app.


MrtCore/winui_desktop_packaged_app/winui_desktop_packaged_app.sln
```diff
@@ -18,6 +18,8 @@ Global
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{28B258CA-5ECC-41EE-A5B0-C2F24F3A4984}.Debug|Any CPU.ActiveCfg = Debug|x86
+		{28B258CA-5ECC-41EE-A5B0-C2F24F3A4984}.Debug|Any CPU.Build.0 = Debug|x86
```
huichen123, line 21: >Debug [](http://example.com/codeflow?start=41&length=5)

also release

huichen123, line 21: >x86 [](http://example.com/codeflow?start=71&length=3)

any cpu?

rohanp-msft: this was auto generated by Visual Studio

huichen123 (approved):
:shipit:

PR #37, opened 2021-05-18
Move the class library into the WinUI C# solution and add a RESW file to the library.

rohanp-msft:
**How verified**
Built and ran the app.
I also updated the README.md file.


MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs
```diff
@@ -0,0 +1,18 @@
...
+// and more about our project templates, see: http://aka.ms/winui-project-info.
+
+namespace ClassLibrary
+{
+    public class Class
+    {
+        public string GetClassLibrarySampleString()
+        {
+            var resourceLoader = new ResourceLoader();
+            var uri = new System.Uri("ms-resource://323858f2-34e1-48a5-8244-5c1bc0c1f889/ClassLibrary/ClassLibraryResources/ClassLibrarySampleString");
```
jefgen, line 14: Just curious, but what is this magic GUID (`323858f2-34e1-48a5-8244-5c1bc0c1f889`)?

rohanp-msft: That's part of the full URI of the resource in the PRI index. It shouldn't be necessary in the URI / string for the lookup, but I couldn't get it to work without it for some reason (I kept getting element not found [for the URI based lookup] or NamedResource not found [for the string based lookup]). @axelandrejs or @huichen123 Do you know why that is?

rohanp-msft: Actually, never mind. Addressed this in iteration 2.

axelandrejs (approved):

MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs
jefgen: Ah, thanks for explaining! :)

jefgen (approved):
:shipit:

huichen123:
> using System.Text;

copyright string

---
Refers to: MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs:1 in 087f350. [](commit_id = 087f350337bd66b241d3c7b4060e14f29cee2e52, deletion_comment = False)

MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Text;
+using Microsoft.ApplicationModel.Resources;
+
+// To learn more about WinUI, the WinUI project structure,
+// and more about our project templates, see: http://aka.ms/winui-project-info.
```
huichen123, line 5: does this have anything to do with WinUI?

MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace ClassLibrary
+{
+    public class Class
+    {
+        public string GetClassLibrarySampleString()
+        {
+            var resourceManager = new ResourceManager();
+            return resourceManager.MainResourceMap.GetValue("ClassLibrary/ClassLibraryResources/ClassLibrarySampleString").ValueAsString;
```
huichen123, line 14: >ClassLibrary/ClassLibraryResources/ClassLibrarySampleString [](http://example.com/codeflow?start=61&length=59)

is this the path in the PRI?

MrtCore/winui_desktop_packaged_app/ClassLibrary/ClassLibrary.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <RootNamespace>ClassLibrary</RootNamespace>
+    <RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Remove="ClassLibraryResources.resw" />
```
huichen123, line 9: >None [](http://example.com/codeflow?start=5&length=4)

does this work for 0.5?

MrtCore/winui_desktop_packaged_app/ClassLibrary/ClassLibrary.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <RootNamespace>ClassLibrary</RootNamespace>
+    <RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Remove="ClassLibraryResources.resw" />
+  </ItemGroup>
```
huichen123, line 10: redundant

MrtCore/winui_desktop_packaged_app/winui_desktop_packaged_app.sln
```diff
@@ -47,6 +52,18 @@ Global
 		{AF3AA2D2-7150-4996-8C38-0AB44604C3D0}.Release|x64.Build.0 = Release|x64
 		{AF3AA2D2-7150-4996-8C38-0AB44604C3D0}.Release|x86.ActiveCfg = Release|x86
 		{AF3AA2D2-7150-4996-8C38-0AB44604C3D0}.Release|x86.Build.0 = Release|x86
+		{E52B115E-E54D-4BF6-8174-4FFF5E8860BE}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{E52B115E-E54D-4BF6-8174-4FFF5E8860BE}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{E52B115E-E54D-4BF6-8174-4FFF5E8860BE}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{E52B115E-E54D-4BF6-8174-4FFF5E8860BE}.Debug|x64.Build.0 = Debug|Any CPU
```
huichen123, line 58: >Any CPU [](http://example.com/codeflow?start=67&length=7)

should these be x64 or anycpu?

MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs
rohanp-msft: removed

MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs
rohanp-msft: Yup:

```
		<ResourceMapSubtree name="ClassLibrary">
			<ResourceMapSubtree name="ClassLibraryResources">
				<NamedResource name="ClassLibrarySampleString" uri="ms-resource://323858f2-34e1-48a5-8244-5c1bc0c1f889/ClassLibrary/ClassLibraryResources/ClassLibrarySampleString">
					<Candidate type="String">
						<Value>English Sample from Class Library</Value>
					</Candidate>
				</NamedResource>
			</ResourceMapSubtree>
		</ResourceMapSubtree>
```

MrtCore/winui_desktop_packaged_app/ClassLibrary/ClassLibrary.csproj
rohanp-msft: If we're adding the file to PRIResource, it makes sense to remove it from None. It belonging to both at the same time doesn't make sense (thinking about it from a Build Action point of view).

Also, this was automatically done. I think it's part of the WinUI project template. If you want, I can remove it.

MrtCore/winui_desktop_packaged_app/ClassLibrary/Class.cs
huichen123: is the format "[project name]/[resw filename]/[string name]"?

MrtCore/winui_desktop_packaged_app/ClassLibrary/ClassLibrary.csproj
huichen123: doesn't include in PRIResource remove it from none automatically? It's strange you have both an include and a remove for the same file.

rohanp-msft: They are separate lists in MSBuild. You have to modify them separately.

MrtCore/winui_desktop_packaged_app/winui_desktop_packaged_app.sln
rohanp-msft: Default for everything in the solution is Any CPU right now. That results in x86 on my dev box, which seems fine to me. If you want me to change the default to x64, I can.

MrtCore/winui_desktop_packaged_app/ClassLibrary/ClassLibrary.csproj
huichen123: what? so a file is in "none" by default unless you explicitly remove it? That's strange. Also does that mean an existing project with PRIResource, and upgrade to Reunion with your build action fix will be broken?

MrtCore/winui_desktop_packaged_app/winui_desktop_packaged_app.sln
huichen123: fine, as long as it has no build warning and runs.

MrtCore/winui_desktop_packaged_app/ClassLibrary/ClassLibrary.csproj
rohanp-msft: All files aren't in None by default. Some might be in Content by default, some in PRIResource by default, etc., depending on the project type.

With my fix, I should have removed the files we added to PRIResource/Content from None. I don't think that affects any functionality, but it's still the right thing to do. I can do that soon.

huichen123: so you mean resw is in None by default? With your fix, If a project has "<PRIResource Include="a.resw" />", will a.resw be added to PRIResource twice?

rohanp-msft: If you copy paste a RESW file into your project folder and add it to the project, it'll belong to the None list.
If you create a new RESW file, it'll belong to the PRIResource list because of the WinUI project template.
With my fix, we only add RESW files from None to PRIResource, so we won't ever have the same RESW file twice in PRIResource.

huichen123: TBH, i don't understand why we need to do remove here.

rohanp-msft: My understanding is that if a file belongs to None, it means that it doesn't belong to any defined group (PRIResource, Content, etc.) So, a file belonging to None and a non-None list at the same time doesn't make sense.

huichen123: does line 16 make it belong to PRIResource?

rohanp-msft: Yes. My point is that that's not sufficient. In MSBuild, None and PRIResource are two separate lists. You have to modify them separately.

huichen123: It turns out your fix already duplicates the file in both lists, and it doesn't seem harmful. I think we shouldn't burden the developer to add this unusual remove.

rohanp-msft: Talked offline. Developer doesn't need to do anything. The include and remove lines are added automatically when they create a new RESW file in a WinUI project. Starting with version 0.8, including preview, all RESW cases will be handled automatically due to this change: https://github.com/microsoft/ProjectReunion/pull/771.

PR #40, opened 2021-05-29
Upgrade all packaged MRT Core sample apps to v0.8 preview.

rohanp-msft:
For the WPF app
1. I removed arm and arm64 configs because arm and arm64 aren't supported:
    - Error: "There was no runtime pack for Microsoft.WindowsDesktop.App.WPF available for the specified RuntimeIdentifier 'win-arm64'."
    - Error: "There was no runtime pack for Microsoft.WindowsDesktop.App.WPF available for the specified RuntimeIdentifier 'win-arm'."

    ARM64 support was added to WPF only in .NET 6 Preview 1: https://github.com/dotnet/wpf/issues/4117

2. I removed AnyCPU config because WinUI doesn't support it.
    - Error: "This version of WinUI does not support AnyCPU. Either set Platform or PlatformTarget to one of the following: x86, x64, or arm64."

For the WinUI C# app
1. I removed AnyCPU config because WinUI doesn't support it.
    - Error: "This version of WinUI does not support AnyCPU. Either set Platform or PlatformTarget to one of the following: x86, x64, or arm64."

**How verified**
Ran all three apps for x86 and x64


jefgen (approved):
:shipit:

PR #56, opened 2021-08-17
Re-calculate drag region for custom TitleBar when window size has changed

rkarman:
The first iteration of the sample lacked recalculating the drag region when the window size changed, this lead to a bug where if you resized the window to a larger size than when you selected to have a custom TitleBar you could not grab the window and move it in the space between the system caption controls and where the drag region was initially calculated to be.

This update breaks out the calculation and setting of the drag region into a separate method and adds an event handler for AppWindow.Changed that calls it when the DidSizeChange property is set, for both the C++ and C# samples.

There are also some minor formatting cleanups in the modified files while I was at it...

### DIFF
--- a/Samples/Windowing/cpp-winui/SampleApp/TitlebarPage.xaml.cpp
+++ b/Samples/Windowing/cpp-winui/SampleApp/TitlebarPage.xaml.cpp
@@ -25,8 +25,17 @@ namespace winrt::SampleApp::implementation
         m_mainWindow = e.Parameter().as<MainWindow>();
         AppWindow appWindow = m_mainWindow.AppWindow();
         m_appWindow = appWindow;
-
+        m_changedToken = m_appWindow.Changed({ this, &TitlebarPage::AppWindowChangedHandler });
     }
+
+    void TitlebarPage::AppWindowChangedHandler(winrt::Windows::Foundation::IInspectable const& sender, winrt::Microsoft::UI::Windowing::AppWindowChangedEventArgs const& e)
+    {
+        if (e.DidSizeChange() && sender.as<AppWindow>().TitleBar().ExtendsContentIntoTitleBar())
+        {
+            SetCustomTitleBarDragRegion();
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


Samples/Windowing/cpp-winui/SampleApp/TitlebarPage.xaml.cpp
```diff
@@ -108,10 +97,37 @@ namespace winrt::SampleApp::implementation
...
+    {
+        // Get the titlebar heigt and set our XAML titlebar element to match
+        int titleBarHeight = m_appWindow.TitleBar().Height();
+        m_mainWindow.MyTitleBar().Height(titleBarHeight);
+
+        // Get caption button occlusion information
+        // Use LeftInset if you've explicitly set your window layout to RTL or if app language is a RTL language
+        double CaptionButtonOcclusionWidth = m_appWindow.TitleBar().RightInset();
+
+        //// Define your drag Regions
```
angelazhangmsft, line 118:
```suggestion
        // Define your drag Regions
```

Samples/Windowing/cs-winui/SampleApp/TitlebarPage.xaml.cs
```diff
@@ -26,19 +26,30 @@ public sealed partial class TitlebarPage : Page
...
             m_mainWindow = e.Parameter as MainWindow;
+            m_mainAppWindow.Changed += MainAppWindow_Changed;
             base.OnNavigatedTo(e);
         }
 
-        public TitlebarPage()
+        private void MainAppWindow_Changed(AppWindow sender, AppWindowChangedEventArgs args)
         {
-            this.InitializeComponent();
+            if(args.DidSizeChange && sender.TitleBar.ExtendsContentIntoTitleBar)
```
angelazhangmsft, line 46:
```suggestion
            if (args.DidSizeChange && sender.TitleBar.ExtendsContentIntoTitleBar)
```

angelazhangmsft (approved):

PR #58, opened 2021-08-19
Display a message box with error code included if dynamic dependencies init fails.

### LINKED ISSUE #48: cs-winforms-unpackaged doesn't check MddBootstrap.Initialize() return code
[WindowsAppSDK-Samples/Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs](https://github.com/microsoft/WindowsAppSDK-Samples/blob/ea9271f18b689228d9251d8a79791a1d3d1170a2/Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs) line 25

```
MddBootstrap.Initialize(8, "preview");
```

doesn't check the return code. Failures will slide by undetected and calls to other APIs will fail (and likely crash) e.g. if the MSIX packages aren't registered WinRT activations will fail due to E_CLASSDB_NOT_REGISTERED. This needs to be something like

```
int hr = MddBootstrap.Initialize(8, "preview");
if (hr < 0)
    …error…
```

rohanp-msft:
Execution terminates after box is dismissed.

Closes #48 

**How verified**
Changed
    `int hr = MddBootstrap.Initialize(8, "preview");`
to
    `int hr = MddBootstrap.Initialize(9, "preview");`
That version doesn't exist.

Dialog I saw:
message_box_2

-2147023727 corresponds to 1169 (0x491) which is "There was no match for the specified key in the index." (see https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--1000-1299-)

### DIFF
--- a/Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
+++ b/Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
@@ -20,9 +20,15 @@ namespace winforms_unpackaged_app
             Application.EnableVisualStyles();
             Application.SetCompatibleTextRenderingDefault(false);
 
-            // Initialize dynamic dependencies so we can consume the Project Reunion APIs in the Project Reunion framework package from this unpackaged app.
+            // Initialize dynamic dependencies so we can consume the Project Reunion APIs in the Project
+            // Reunion framework package from this unpackaged app.
             // Take a dependency on Project Reunion v0.8 preview.
-            MddBootstrap.Initialize(8, "preview");
+            int hr = MddBootstrap.Initialize(8, "preview");
+            if (hr < 0)
+            {
+                MessageBox.Show("Fail to initialize Dynamic Dependencies with error " + hr, "Error!");
+                return;
+            }
             
             var resourceManager = new ResourceManager("winforms_unpackaged_app.pri");
             // Fall back to other resource loaders if the resource is not found in MRT, in this case .net.
--- a/Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
+++ b/Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
@@ -26,7 +26,7 @@ namespace winforms_unpackaged_app
             int hr = MddBootstrap.Initialize(8, "preview");
             if (hr < 0)
             {
-                MessageBox.Show("Fail to initialize Dynamic Dependencies with error " + hr, "Error!");
+                MessageBox.Show("Fail to initialize Dynamic Dependencies with error 0x" + hr.ToString("X"), "Error!");
                 return;
             }
             


Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
```diff
@@ -1,54 +1,60 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
jefgen, line 1: Curiously the whole file shows as changed in the diff on GitHub... I wonder if maybe the encoding of the file changed perhaps?

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
```diff
@@ -1,54 +1,60 @@
...
+        ///  The main entry point for the application.
+        /// </summary>
+        [STAThread]
+        static void Main()
+        {
+            Application.SetHighDpiMode(HighDpiMode.SystemAware);
+            Application.EnableVisualStyles();
+            Application.SetCompatibleTextRenderingDefault(false);
+
+            // Initialize dynamic dependencies so we can consume the Project Reunion APIs in the Project
```
jefgen, line 23:
```suggestion
            // Initialize dynamic dependencies so we can consume the Windows App SDK APIs in the Project

```

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
rohanp-msft: VS asked if I wanted to fix line endings and I said yes. That may be why. In CF, only my additions show as new...

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
```diff
@@ -1,54 +1,60 @@
...
+            var overrideResourceContext = resourceManager.CreateResourceContext();
+            overrideResourceContext.QualifierValues["Language"] = "de-DE";
+
+            // The resource manager and the resource loader do disk I/O and should be centrally cached.
+            // Since both end up loading the same file, ideally an app should only use one.
+            // The resource manager is a superset of the resource loader in terms of functionality.
+            Application.Run(new AppForm(resourceManager, resourceLoader, overrideResourceContext));
+
+            // Uninitialize dynamic dependencies.
+            MddBootstrap.Shutdown();
```
axelandrejs, line 57: @DrusTheAxe it occurred to me that requiring an explicit shutdown call is prone to leak in exception-based languages. Do you have recommendation for handling this? Does this object unload on scope exit, or does it have a RAII wrapper?

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
rohanp-msft: This sample uses 0.8 Preview where the name is Project Reunion. Soon, the main branch will be forked to create the 0.8 branch. When we then update the apps in main to use some version of 1.0, we should update the name as well.

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
```diff
@@ -1,54 +1,60 @@
...
+        static void Main()
+        {
+            Application.SetHighDpiMode(HighDpiMode.SystemAware);
+            Application.EnableVisualStyles();
+            Application.SetCompatibleTextRenderingDefault(false);
+
+            // Initialize dynamic dependencies so we can consume the Project Reunion APIs in the Project
+            // Reunion framework package from this unpackaged app.
+            // Take a dependency on Project Reunion v0.8 preview.
+            int hr = MddBootstrap.Initialize(8, "preview");
```
axelandrejs, line 26: Why does this method return an HRESULT in C#? Shouldn't this project to an exception?

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
jefgen: Ah, got it. Thanks for explaining!

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
jefgen: I think the `MddBootstrap` object is part of the sample program itself: https://github.com/microsoft/WindowsAppSDK-Samples/blob/e6d6bd953a6498b9efd510de80591edd862ae8f8/Samples/ResourceManagement/cs-winforms-unpackaged/MddBootstrap.cs#L49

Maybe/Perhaps that class could be updated to be more of a RAII object and call `Shutdown` when destructed?

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
jefgen: Ah, from offline discussion what is happening is that the file is being changed from `LF` endings to `CR` `LF` endings.
Since CR LF is more common/expected on Windows, it is probably better to change this anyways...

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
axelandrejs: Good catch! Yes, the class should be RAII-ified. In fact we should probably create the canonical C# wrapper for this, and a canonical C++ RAII wrapper for this. Otherwise every app will create its own.

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
rohanp-msft: @DrusTheAxe  also.

Would `Initialize()` throwing an exception and Program.cs catching it and displaying the message box be the C# way?

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
```diff
@@ -1,54 +1,60 @@
...
+            if (hr < 0)
+            {
+                MessageBox.Show("Fail to initialize Dynamic Dependencies with error " + hr, "Error!");
```
huichen123, line 29: >hr [](http://example.com/codeflow?start=88&length=2)

is this in decimal?

huichen123: see your description. few people can read decimal HRESULT. make it hex.

rohanp-msft: done!

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
axelandrejs: Yes!

DrusTheAxe: <del>Winter</del>Bootstrap C# API is coming...

[Spec updates for GenerationId and Bootstrap C# API #1284](https://github.com/microsoft/WindowsAppSDK/pull/1284)

[Add Bootstrap C# API. Spec Bootstrap C# and get GenerationId (Flat-C/WinRT) APIs #1274](https://github.com/microsoft/WindowsAppSDK/pull/1274)

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
DrusTheAxe: MddBootstrap.cs was a quick helper thrown together for initial samples. The Bootstrap C# API is on its way...

[Spec updates for GenerationId and Bootstrap C# API #1284](https://github.com/microsoft/WindowsAppSDK/pull/1284)

[Add Bootstrap C# API. Spec Bootstrap C# and get GenerationId (Flat-C/WinRT) APIs #1274](https://github.com/microsoft/WindowsAppSDK/pull/1274)

Once complete samples can (should) use the API and delete their copy of MddBootstrap.cs

DrusTheAxe: > canonical C++ RAII wrapper

A C++ RAII wrapper isn't planned as the value seemed low (and in some ways, negative). Given WIL, `RETURN_IF_FAILED(MddBootstrapInitialize(...))` or `THROW_IF_FAILED(MddBootstrapInitialize(...))` didn't see a problem.

As for RAII, that is questionable and potentially dangerous. `MddBootstrapShutdown()` reverts what `MddBootstrapInitialize ()` did, which is mainly memory fiddling. As this is expected to be done once at the process level (e.g. in main/WinMain) then unfiddling memory right before the process ends isn't important. The key thing MddBootstrapInitialize does is to create a running process with the identity of a DDLM package (to ensure Windows doesn't service the WinAppSDK Framework while you're using it). That process is kept running until `MddBootstrapShutdown`, which ends it.

Today that process is a PackagedCOM OOP server, so shutdown is `p->Release()` to explicitly stop using it (and if this was the only process with a reference, to end the process). If you don't call it and just ExitProcess() Windows will cleanup your process' references to OOP COM servers. Calling `MddBootstrapShutdown` is slightly more efficient than ending your process without calling it, but same net result.

So, no harm no foul if you don't call `MddBootstrapShudown()` when your process is ending.

Given a RAII wrapper `MddBootstrapShutdown` would be called by the object's destructor when the object goes out of scope. Which sounds good...unless you have anything using WinAppSDK executing after that. That could be harmless to problematic to fatal, and easily can lead to subtle and cryptic problems. So, maybe harmless, maybe not so.

Hence the reason the Bootstrap API was provided as a C API for C++ developers.

Sure it's possible we could provide a C++ RAII wrapper. What benefit do you think that would provide? And does that outweigh the potential downsides?

NOTE: The Bootstrap C# API exists for multiple reasons (some of which don't apply to C++). See the links above for more details.

DrusTheAxe: > requiring an explicit shutdown call is prone to leak

Usually...but not here.

Failure to call MddBootstrapShutdown() at process termination means Windows' process cleanup will drop the reference to the DDLM OOP COM server. No leak, just a slight difference in timing and codepath.

It also means we won't unfiddle the memory fiddled with by `MddBootstrapInitialize` (altered PATH environment variable, etc). But if the process is terminating then that's harmless.

And the Bootstrap API is designed to be called once before/after all WinAppSDK usage in a process. In practice that means at the very start and end of `main()`/`WinMain()`.

I'm a huge RAII supporter. Call this the exception providing the rule

(though I'm open to feedback if folks think a C++ RAII wrapper is helpful)

DrusTheAxe (approved):

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
rohanp-msft: I see that the Bootstrap C# API Howard references above includes a throwing `Initialize()`. So I'll wait for that.

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
rohanp-msft: Given Howard's argument, I'll leave the class as it is. Further discussion and a decision may be pending, but that can be done separately and any action needed can be taken separately.

Samples/ResourceManagement/cs-winforms-unpackaged/Program.cs
rohanp-msft: Keeping the new line endings per Jeff's reasoning.

PR #69, opened 2021-08-31
Add sample WinUI C#/C++ templates

angelazhangmsft:
Outstanding issues
- C++ SettingsPage theme changes 
- How to reference all Assets (images) from a shared folder for C#


SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,77 @@
+﻿#include "pch.h"
```
angelazhangmsft, line 1: @marb2000 @MikeHillberg the settings theme changes here aren't working currently. Would you be able to review this code?

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+        GetWindowRect(hwnd, &rc);
+        ClipOrCenterRectToMonitorWin32(&rc, true);
+        SetWindowPos(hwnd, HWND_TOP,
+            rc.left, rc.top, 0, 0,
+            SWP_NOSIZE |
+            SWP_NOZORDER |
+            SWP_NOACTIVATE);
+    }
+
+    void MainWindow::ClipOrCenterRectToMonitorWin32(RECT* prc, bool UseWorkArea)
```
jonwis, line 81: Consistency - parameters should be `camelCase`

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+#include "pch.h"
+
+namespace winrt::WinUI3TemplateCpp
+{
+	class Settings
+	{
+	public:
+		static hstring FeatureName;
```
jonwis, line 9: Indentation consistency - use spaces, not tabs here, and only 4 spaces after a `{`

SharedContent/cpp-winui-template/WinUI3TemplateCpp/Scenario1_ShortName.xaml.cpp
```diff
@@ -0,0 +1,46 @@
...
+    WinUI3TemplateCpp::MainPage Scenario1_ShortName::rootPage{ nullptr };
+
+    Scenario1_ShortName::Scenario1_ShortName()
+    {
+        InitializeComponent();
+        Scenario1_ShortName::rootPage = WinUI3TemplateCpp::MainPage::Current();
+    }
+
+    
+    void Scenario1_ShortName::SuccessMessage_Click(winrt::Windows::Foundation::IInspectable const& sender, winrt::Microsoft::UI::Xaml::RoutedEventArgs const& e)
```
jonwis, line 24: Consider removing the parameter names, or you're going to get VC complaining about "unreferenced formal parameter". So say this:

```c++
void Scenario1_ShortName::SuccessMessage_Click(winrt::Windows::Foundation::IInspectable const&, RoutedEventArgs const&)
```

SharedContent/cpp-winui-template/WinUI3TemplateCpp/Scenario1_ShortName.xaml.cpp
```diff
@@ -0,0 +1,46 @@
+﻿#include "pch.h"
+#include "Scenario1_ShortName.xaml.h"
+#if __has_include("Scenario1_ShortName.g.cpp")
+#include "Scenario1_ShortName.g.cpp"
+#endif
+
+using namespace winrt;
+using namespace Microsoft::UI::Xaml;
```
jonwis, line 8: Consider adding `using namespace Windows::Foundation;` as well, so you can just say `IInspectable const&`

SharedContent/cpp-winui-template/WinUI3TemplateCpp/Scenario1_ShortName.xaml.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+    void Scenario1_ShortName::InformationalMessage_Click(winrt::Windows::Foundation::IInspectable const& sender, winrt::Microsoft::UI::Xaml::RoutedEventArgs const& e)
+    {
+        rootPage.NotifyUser(L"Everything was ok!", InfoBarSeverity::Informational, true);
+    }
+
+
+    void Scenario1_ShortName::ClearMessage_Click(winrt::Windows::Foundation::IInspectable const& sender, winrt::Microsoft::UI::Xaml::RoutedEventArgs const& e)
+    {
+        rootPage.NotifyUser(L"Everything was ok!", InfoBarSeverity::Informational, false);
```
jonwis, line 44: Looks like these repeat the same notify string, even for the "error" one?

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+    SettingsPage::SettingsPage()
+    {
+        InitializeComponent();
+    }
+
+    void SettingsPage::OnNavigatedTo(Microsoft::UI::Xaml::Navigation::NavigationEventArgs const& e)
+    {
+        // Initialize the CurrentTheme on first navigation to the Settings page
+        if (winrt::WinUI3TemplateCpp::Settings::CurrentTheme == L"")
```
jonwis, line 22: Consistency - below you don't repeat the `winrt::WinUI3TemplateCpp` namespace on `Settings::X`; choose one

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,77 @@
...
+                RadioButton radioButton = c.as<RadioButton>();
+                radioButton.IsChecked(Windows::Foundation::IReference<bool>{true});
+            }
+        }
+    }
+
+    void SettingsPage::OnThemeRadioButtonChecked(winrt::Windows::Foundation::IInspectable const& sender, winrt::Microsoft::UI::Xaml::RoutedEventArgs const& e)
+    { 
+        RadioButton radiobutton = sender.as<RadioButton>();
+		winrt::hstring selectedTheme = unbox_value<winrt::hstring>(radiobutton.Tag());
```
jonwis, line 52: Indentation again. Maybe @chrisguzak could help you add a `.editorconfig` that would ensure consistent formatting?

SharedContent/cs-winui-template/WinUI3TemplateCs/Scenario1_ShortName.xaml.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            this.InitializeComponent();
+        }
+
+        private void SuccessMessage_Click(object sender, RoutedEventArgs e)
+        {
+            rootPage.NotifyUser("Everything was ok!",InfoBarSeverity.Success);
+        }
+
+        private void ErrorMessage_Click(object sender, RoutedEventArgs e)
```
jonwis, line 22: Looks like these strings are fixed, though, so that's good.

jonwis (changes requested):
Good start - some formatting comments and a string or two to check.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include "SampleConfiguration.h"
+#include "MainPage.xaml.h" 
+
+using namespace winrt;
+using namespace winrt::WinUI3TemplateCpp;
+using namespace Windows::Foundation::Collections;
+
+IVector<Scenario> implementation::MainPage::scenariosInner = winrt::single_threaded_observable_vector<Scenario>(
+	{
+		Scenario{ L"Scenario 1", L"WinUI3TemplateCpp.Scenario1_ShortName"},
```
oldnewthing, line 12:
```suggestion
		Scenario{ L"Scenario 1", winrt::name_of<WinUI3TemplateCpp::Scenario1_ShortName>() },
```

Use `name_of` so that refactoring will update the page name when the class is renamed.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml
```diff
@@ -0,0 +1,40 @@
...
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:WinUI3TemplateCpp"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <Grid RowDefinitions="*,Auto">
+
+        <StackPanel HorizontalAlignment="Left" Spacing="20" >
```
oldnewthing, line 12: Should make an XML prettify pass over these XAML files to ensure consistency of indentation, attribute alignment, space-before-close-tag, etc.

Should also have a comment in the samples guide about what prettify parameters to use.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml
```diff
@@ -0,0 +1,40 @@
...
+            <TextBlock  Style="{StaticResource TitleTextBlockStyle}"
+                                Margin="0,4,0,0"
+                                FontWeight="Normal"
+                                Text="Theme Mode" />
+            <StackPanel x:Name="themePanel">
+                <RadioButton Tag="Light" Checked="OnThemeRadioButtonChecked" Content="Light"  />
```
oldnewthing, line 18: Don't these radio buttons need a GroupName?

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    void SettingsPage::OnNavigatedTo(Microsoft::UI::Xaml::Navigation::NavigationEventArgs const& e)
+    {
+        // Initialize the CurrentTheme on first navigation to the Settings page
+        if (winrt::WinUI3TemplateCpp::Settings::CurrentTheme == L"")
+        {
+            if (RequestedTheme() == ElementTheme::Dark) 
+            {
+                Settings::CurrentTheme = L"Dark";
```
oldnewthing, line 26: Not a fan of using a string to record the theme. Can we just make the tag an ElementTheme?
```xml
<RadioButton Checked="OnThemeRadioButtonChecked" Content="Dark">
    <RadioButton.Tag><ElementTheme>Dark</ElementTheme></RadioButton.Tag>
</RadioButton>
```

Then we would do `c.as<RadioButton>().Tag().as<ElementTheme>()`.

If we like strings, we should at least convert the string to an ElementTheme immediately upon parsing the tag, and then work in ElementTheme from then on.

```cpp
ElementTheme ParseElementTheme(hstring const& s)
{
    if (s == L"Light") return ElementTheme::Light;
    if (s == L"Dark") return ElementTheme::Dark;
    return ElementTheme::Default;
}
```

SharedContent/cpp-winui-template/WinUI3TemplateCpp/App.xaml.h
```diff
@@ -0,0 +1,21 @@
...
+    struct App : AppT<App>
+    {
+        App();
+
+        void OnLaunched(Microsoft::UI::Xaml::LaunchActivatedEventArgs const&);
+        
+    private:
+        winrt::Microsoft::UI::Xaml::Window window{ nullptr };
+    };
+}
```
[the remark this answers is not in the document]
angelazhangmsft: For the upcoming 1.x releases, probably ~10 features (which may have multiple samples per feature). I think "Templates" would be a better name, since not all samples will use these templates (for example, we also will have Win32/WPF/WinForms/Console app samples). 

Having to edit all WinUI samples whenever updating the template may be a nuisance, but we might need to do this for early releases of WinAppSDK

SharedContent/cs-winui-template/WinUI3TemplateCs/MainWindow.xaml.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        private unsafe void LoadIcon(HWND hwnd, string iconName)
+        {
+            const int ICON_SMALL = 0;
+            const int ICON_BIG = 1;
+
+            fixed (char* nameLocal = iconName)
+            {
+                HINSTANCE hInstLocal = default(HINSTANCE);
+                HANDLE __result = LoadImage(hInstLocal, nameLocal,
```
AArnott, line 37: rename `__result` to something more meaningful to the user. It had that odd prefix and generic name in generated code, but in user-owned code it should be something better.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainWindow.xaml.cs
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        private unsafe void LoadIcon(HWND hwnd, string iconName)
+        {
+            const int ICON_SMALL = 0;
+            const int ICON_BIG = 1;
+
+            fixed (char* nameLocal = iconName)
+            {
+                HINSTANCE hInstLocal = default(HINSTANCE);
```
AArnott, line 36: No need for a local declaration here. You can just use `default(HINSTANCE)` as the first argument.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainWindow.xaml.cs
```diff
@@ -0,0 +1,110 @@
...
+        private void ClipOrCenterRectToMonitorWin32(RECT prc, bool UseWorkArea, bool IsCenter)
+        {
+            HMONITOR hMonitor;
+            RECT rc;
+            int w = prc.right - prc.left;
+            int h = prc.bottom - prc.top;
+
+            hMonitor = MonitorFromRect(prc, MONITOR_FROM_FLAGS.MONITOR_DEFAULTTONEAREST);
+            MONITORINFO mi = new MONITORINFO();
+            mi.cbSize = (uint)Marshal.SizeOf(mi);
```
AArnott, line 88: Use `sizeof(MONITORINFO)` instead of `Marshal.SizeOf` if C# will let you. It results in a constant instead of a method call.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+namespace WinUI3TemplateCs
+{
+
+    public sealed partial class MainPage : Page
+    {
+        public static MainPage Current;
+
+        public List<Scenario> Scenarios
+        {
+            get { return this.scenarios; }
```
AArnott, line 18: This could be written more succinctly as `get => this.scenarios;`
Or even better, replace the whole property with:

```cs
public List<Scenario> Scenarios => this.scenarios`
```

Or consider removing the field altogether and assigning the property directly:

```cs
public List<Scenario> Scenarios { get; } = new()
{
    new Scenario() { Title = "Scenario 1", ClassName = typeof(Scenario1_ShortName).FullName },
    new Scenario() { Title = "Scenario 2", ClassName = typeof(Scenario2_ShortName).FullName }
};
```

And if the collection is not expected to change at runtime, consider exposing as an `IReadOnlyList<Scenario>` instead.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+            // here to load the home page.
+            if (scenarios is not null && scenarios.Count > 0)
+            {
+                NavView_Navigate(scenarios.First().ClassName, new Microsoft.UI.Xaml.Media.Animation.EntranceNavigationTransitionInfo());
+            }
+        }
+
+        private void NavView_Navigate(string navItemTag, Microsoft.UI.Xaml.Media.Animation.NavigationTransitionInfo transitionInfo)
+        {
+            Type _page = null;
```
AArnott, line 80: IMO leading underscores are ugly in C#. But where they are used, there is a strong convention for reserving them for *field* names to distinguish them from local variables. Either way, this local shouldn't be using a leading underscore.
Also, avoid initializing a local that should be set to a value later. In this case, both branches of the next `if` statement assign this local. If a bug were to creep in where it was no longer assigned, the C# compiler could help call it out -- but only if you *don't* assign it here.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            if (navItemTag == nameof(SettingsPage))
+            {
+                _page = typeof(SettingsPage);
+            }
+            else
+            {
+                var item = scenarios.FirstOrDefault(p => p.ClassName.Equals(navItemTag));
```
AArnott, line 88: tip: avoid `var` where the type is not obvious elsewhere in the statement.

Here and everywhere else.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+                _page = Type.GetType(item.ClassName);
```
AArnott, line 89: You are dereferencing `item` without a null check, but you just used `FirstOrDefault` which explicitly allows for returning `null`. This is a `NullReferenceException` risk. If you really expect a match on the previous line, use `First` instead so it throws a more diagnosable exception rather than returning null. Otherwise add a null check.

This is the sort of bug that can be avoided by adding `<Nullable>enable</Nullable>` to the project file.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+            }
+
+            // Get the page type before navigation so you can prevent duplicate
+            // entries in the backstack.
+            var preNavPageType = ContentFrame.CurrentSourcePageType;
+
+            // Only navigate if the selected page isn't currently loaded.
+            if (!(_page is null) && !Type.Equals(preNavPageType, _page))
```
AArnott, line 97: replace `!(_page is null)` with `_page is not null` for readability.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                ContentFrame.GoBack();
+            }
+        }
+
+        private void NavView_ItemInvoked(NavigationView sender, NavigationViewItemInvokedEventArgs args)
+        {
+            var naViewItemInvoked = (NavigationViewItem)args.InvokedItemContainer;
+
+            if (args.IsSettingsInvoked == true)
```
AArnott, line 115: nit: drop the `== true`. That is entirely redundant for a `bool` returning property.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                NavView_Navigate(nameof(SettingsPage), args.RecommendedNavigationTransitionInfo);
+            }
+            else
+            if (args.InvokedItemContainer is not null)
```
AArnott, line 120: nit: keep `else` and `if` on the same line.

Here and elsewhere.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                var navItemTag = args.InvokedItemContainer.Tag?.ToString();
+                if (navItemTag is not null)
```
AArnott, line 123: Considering its source, `!string.IsNullOrEmpty(navItemTag)` may be the more appropriate check here.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+            if (ContentFrame.SourcePageType == typeof(SettingsPage))
+            {
+                // SettingsItem is not part of NavView.MenuItems, and doesn't have a Tag.
+                NavView.SelectedItem = (NavigationViewItem)NavView.SettingsItem;
+                NavView.Header = "Settings";
+            }
+            else
+            if (ContentFrame.SourcePageType != null)
+            {
+                var item = scenarios.FirstOrDefault(p => p.ClassName == e.SourcePageType.FullName);
```
AArnott, line 143: Another use of `FirstOrDefault` followed by dereferencing of the result without a null check.

SharedContent/cs-winui-template/WinUI3TemplateCs/App.xaml.cs
```diff
@@ -0,0 +1,21 @@
...
+            this.InitializeComponent();
+        }
+
+        protected override void OnLaunched(LaunchActivatedEventArgs args)
+        {
+            m_window = new MainWindow();
+            m_window.Activate();
+        }
+
+        private Window m_window;
```
AArnott, line 19: nit: fields should be on top of the class, above the constructor.  See [SA1201](https://documentation.help/StyleCop/SA1201.html).

AArnott: Also: we don't use hungarian prefixes in C#. You don't anywhere else in this project either. Can you name this `mainWindow`?

SharedContent/cs-winui-template/WinUI3TemplateCs/App.xaml
```diff
@@ -0,0 +1,135 @@
...
+            </Style>
+
+            <Style x:Key="ScenarioDescriptionTextStyle" TargetType="TextBlock" BasedOn="{StaticResource BodyTextBlockStyle}">
+            </Style>
+
+            <Style x:Key="BaseMessageStyle" TargetType="TextBlock" BasedOn="{StaticResource BodyTextBlockStyle}">
+                <Setter Property="Margin" Value="0,0,0,5"/>
+            </Style>
+
+
```
AArnott, line 132: Avoid multiple blank lines in a row.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            // If called from the UI thread, then update immediately.
+            // Otherwise, schedule a task on the UI thread to perform the update.
+            if (DispatcherQueue.HasThreadAccess)
+            {
+                UpdateStatus(strMessage, severity, isOpen);
+            }
+            else
+            {
+                DispatcherQueue.TryEnqueue(() =>
```
AArnott, line 40: Your pattern here requires repeating the same code in two blocks. While it's only one line (for now), it's an unfortunate pattern that may be copied where it's even worse.

I wonder how folks would feel about using `JoinableTaskFactory` in this template. If you did, this entire method body could be replaced with just this:

```cs
await App.JoinableTaskFactory.SwitchToMainThreadAsync();
UpdateStatus(strMessage, severity, isOpen);
```

AArnott:
Mostly style feedback, but a few functional concerns/ideas as well.

AArnott (commented):
Mostly style feedback, but a few functional concerns/ideas as well.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml
AArnott: Adding an `.editorconfig` file to the sln that captures the coding style you're following would be good too.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+        LoadIcon(hwnd, L"windows-sdk.ico");
+        SetWindowSize(hwnd, 1220, 1080);
+        PlacementCenterWindowInMonitorWin32(hwnd);
+    }
+
+    HWND MainWindow::GetWindowHandle()
+    {
+        if (_hwnd == nullptr)
+        {
+            Window window = this->try_as<Window>();
```
AArnott, line 32: What does `try_as` do when it fails? If it returns a null reference, the app may crash on the next line with something less helpful than a cast syntax that throws when it fails.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+            GetSystemMetrics(SM_CXICON),
+            GetSystemMetrics(SM_CYICON),
+            LR_LOADFROMFILE);
+        SendMessage(hwnd, WM_SETICON, ICON_BIG, (LPARAM)hBigIcon);
+    }
+
+    void MainWindow::SetWindowSize(const HWND hwnd, const int width, const int height)
+    {
+        // Win32 uses pixels and WinUI 3 uses effective pixels, so you should apply the DPI scale factor
+        auto dpi = GetDpiForWindow(hwnd);
```
AArnott, line 61: Consider avoiding `auto` when the type isn't obvious. Particularly in light of the cast bug below, seeing that `dpi` is an integer rather than a `float` may help avoid such bugs.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+            LR_LOADFROMFILE);
+        SendMessage(hwnd, WM_SETICON, ICON_SMALL, (LPARAM)hSmallIcon);
+
+        auto hBigIcon = LoadImage(
+            NULL,
+            iconName,
+            IMAGE_ICON,
+            GetSystemMetrics(SM_CXICON),
+            GetSystemMetrics(SM_CYICON),
+            LR_LOADFROMFILE);
```
AArnott, line 54: Same feedback as I left on the C# side around a missing `LR_SHARED` flag here.

SharedContent/cs-winui-template/WinUI3TemplateCs/SampleConfiguration.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public static class Settings
+    {
+        public const string FeatureName = "Windows App SDK Template C# Sample";
+        public static string CurrentTheme;
+    }
+
+    public partial class MainPage : Page
+    {
+        private readonly List<Scenario> scenarios = new()
```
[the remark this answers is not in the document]
oldnewthing: This pattern was inherited from the UWP samples repo, where the `MainPage.xaml.cs` file is shared among all samples to save disk space and make it easier to update all 200+ samples at one go (by updating the shared copy). Under that model, sample-specific stuff goes into `SampleConfiguration.cs`.

If we intend to follow the same pattern here, then the scenario list needs to live somewhere outside the shared copy (since it changes from sample to sample).

Even if we decide that each sample gets its own copy of `MainPage.xaml.cs`, we might want to keep it separate anyway, because that would make it easier to update all the samples at once (by replacing the MainPage.xaml.cs with an updated version).

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
oldnewthing: This was probably inherited from the UWP samples template, which predates many of the C# conciseness features people have become accustomed to.

At least in the UWP side, there are rare cases where the list of scenarios is edited on the fly. (E.g., some scenarios are not available when running on Phone or Xbox.)

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
oldnewthing: If you are going to assume that the result is never null, then use `First()` rather than `FirstOrDefault()`. `First()` requires that a match be found. `FirstOrDefault()` says "I'll look for one, but if I can't find one, I'll return null." Don't ask for a feature that makes you crash. Let `First` report the error. That makes the problem much more diagnosable.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
oldnewthing: +1. Use `as<>` when you require a non-null result. Use `try_as<>` if you want to handle failure.

I've found this to be a popular anti-pattern: Using a special version that returns null, and then crashing on null anyway. If you are going to crash on null, then don't ask for the special version that returns null.

SharedContent/README.md
```diff
@@ -0,0 +1,9 @@
+#  Windows App SDK Sample Templates
+
+The `SharedContents` folder contains sample templates for C# and C++ (using WinUI 3). To use the current templates, sample authors should follow these steps:
+
```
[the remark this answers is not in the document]
angelazhangmsft: Created PR adding the file to this repo

SharedContent/cpp-winui-template/WinUI3TemplateCpp/App.xaml.cpp
```diff
@@ -0,0 +1,34 @@
...
+using namespace Microsoft::UI::Xaml::Navigation;
+using namespace WinUI3TemplateCpp;
+using namespace WinUI3TemplateCpp::implementation;
+
+App::App()
+{
+    InitializeComponent();
+
+#if defined _DEBUG && !defined DISABLE_XAML_GENERATED_BREAK_ON_UNHANDLED_EXCEPTION
+    UnhandledException([this](IInspectable const&, UnhandledExceptionEventArgs const& e)
```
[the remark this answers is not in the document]
angelazhangmsft: What is this correct pattern here? Removing `[this]` results in "typename not allowed errors"

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+        SendMessage(hwnd, WM_SETICON, ICON_BIG, (LPARAM)hBigIcon);
+    }
+
+    void MainWindow::SetWindowSize(const HWND hwnd, const int width, const int height)
+    {
+        // Win32 uses pixels and WinUI 3 uses effective pixels, so you should apply the DPI scale factor
+        auto dpi = GetDpiForWindow(hwnd);
+        float scalingFactor = static_cast<float>(dpi / 96);
```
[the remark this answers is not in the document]
angelazhangmsft: I am seeing strange behavior with the window sizing when casting to float first. The app will open up in a window that is taller than my screen.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainPage.xaml.cpp
```diff
@@ -0,0 +1,155 @@
...
+    {
+        // If called from the UI thread, then update immediately.
+        // Otherwise, schedule a task on the UI thread to perform the update.
+        if (this->DispatcherQueue().HasThreadAccess())
+        {
+            UpdateStatus(strMessage, severity, isOpen);
+        }
+        else
+        {
+            this->DispatcherQueue().TryEnqueue([strMessage, severity, isOpen, this]()
```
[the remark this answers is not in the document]
angelazhangmsft: From offline, wil::resume_foreground might not be available. May stick with the current pattern for now, since we don't care if the DQ has shut down.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/Scenario1_ShortName.xaml.cpp
angelazhangmsft: Yes, the strings were repeated by accident - this one should be an empty string. Makes sense to remove all instances of `isOpen` in both C#/C++ templates and replace with:

```cpp
if (strMessage == L"")
{
    infoBar().IsOpen(false);
}
else 
{
    infoBar().IsOpen(true);
}  
```

SharedContent/README.md
```diff
@@ -0,0 +1,9 @@
...
+1. Clone the Windows App SDK samples repo and then copy all the files under the appropriate template folder to your feature sample directory. For example, create a new folder for your sample called `Samples/MyFeatureName/cs-winui/`, and copy all the contents under `SharedContent/cs-winui-template` to this new folder.
```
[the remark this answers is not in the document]
angelazhangmsft: It would be good to have a VSIX eventually, but we will need to figure out how to update these when new versions of WinAppSDK are released. Is there a way to generate a unique identity without a VSIX?

SharedContent/cpp-winui-template/WinUI3TemplateCpp/Scenario1_ShortName.xaml.cpp
angelazhangmsft: removed all parameter names that cause "unreferenced formal parameter" warnings

SharedContent/cpp-winui-template/WinUI3TemplateCpp/App.xaml.h
```diff
@@ -0,0 +1,21 @@
...
+namespace winrt::WinUI3TemplateCpp::implementation
+{
+    struct App : AppT<App>
+    {
+        App();
+
+        void OnLaunched(Microsoft::UI::Xaml::LaunchActivatedEventArgs const&);
+        
+    private:
+        winrt::Microsoft::UI::Xaml::Window window{ nullptr };
```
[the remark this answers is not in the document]
angelazhangmsft: removed `winrt` namespace qualifer

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainPage.xaml.cpp
```diff
@@ -0,0 +1,155 @@
...
+        InitializeComponent();
+        MainPage::current = *this;
+    }
+
+    void MainPage::NavView_Loaded(winrt::Windows::Foundation::IInspectable const& sender, winrt::Microsoft::UI::Xaml::RoutedEventArgs const& e)
+    {
+        for (auto s : Scenarios())
+        {
+            FontIcon fontIcon{};
+            fontIcon.FontFamily(winrt::Microsoft::UI::Xaml::Media::FontFamily(L"Segoe MDL2 Assets"));
```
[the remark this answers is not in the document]
angelazhangmsft: Added namespace composition to MainPage.xaml.cpp. Is it ok that only this file uses namespace composition but the other files don't (consistency)?

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainPage.xaml.cpp
```diff
@@ -0,0 +1,155 @@
...
+            NavView().MenuItems().Append(navViewItem);
+        }
+
+        // NavView doesn't load any page by default, so load home page.
+        NavView().SelectedItem(NavView().MenuItems().GetAt(0));
+
+        // If navigation occurs on SelectionChanged, this isn't needed.
+        // Because we use ItemInvoked to navigate, we need to call Navigate
+        // here to load the home page.
+        if (Scenarios() != nullptr && Scenarios().Size() > 0)
```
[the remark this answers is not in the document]
angelazhangmsft: Removed the first condition testing for `nullptr`.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+    {
+        if (_hwnd == nullptr)
+        {
+            Window window = this->try_as<Window>();
+            window.as<IWindowNative>()->get_WindowHandle(&_hwnd);
+        }
+        return _hwnd;
+    }
+
+    void MainWindow::LoadIcon(HWND hwnd, wchar_t* iconName)
```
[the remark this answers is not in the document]
angelazhangmsft: Good suggestion, changed

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    void MainWindow::SetWindowSize(const HWND hwnd, const int width, const int height)
+    {
+        // Win32 uses pixels and WinUI 3 uses effective pixels, so you should apply the DPI scale factor
+        auto dpi = GetDpiForWindow(hwnd);
+        float scalingFactor = static_cast<float>(dpi / 96);
+        auto widthScaled = static_cast<int>(width * scalingFactor);
+        auto heightScaled = static_cast<int>(height * scalingFactor);
+
+        SetWindowPos(hwnd, HWND_TOP, 0, 0, widthScaled, heightScaled, SWP_NOMOVE);
```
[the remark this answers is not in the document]
angelazhangmsft: replaced `HWND_TOP` with `nullptr` and `SWP_NOMOVE` to `SWP_NOZORDER`

SharedContent/cpp-winui-template/WinUI3TemplateCpp/App.xaml.h
```diff
@@ -0,0 +1,21 @@
+﻿#pragma once
+
+#pragma push_macro("GetCurrentTime")
+#undef GetCurrentTime
```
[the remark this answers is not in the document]
angelazhangmsft: removed lines 3/4 here.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.cpp
angelazhangmsft: changed to using `name_of`

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/Package.appxmanifest
```diff
@@ -0,0 +1,52 @@
...
+  IgnorableNamespaces="uap rescap">
+
+  <Identity
+    Name="cfb926cc-c146-49f1-9ac1-cfaf8182bfd7"
+    Publisher="CN=Microsoft Corporation"
+    Version="1.0.0.0" />
+
+  <Properties>
+    <DisplayName>WinUICSTemplate (Package)</DisplayName>
+    <PublisherDisplayName>Microsoft Corporation</PublisherDisplayName>
```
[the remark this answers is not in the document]
angelazhangmsft: I think since these are templates for Microsoft samples, this is ok. The samples are mainly for showcasing how to use specific APIs. I followed the existing UWP sample templates: https://github.com/microsoft/Windows-universal-samples/blob/ca74baf5c9e53955e8d749733b8e3e4839afd472/SharedContent/Templates/UWPSDKSampleCS/Package.appxmanifest#L9

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
angelazhangmsft: @marb2000 do you have thoughts on switching from `DispatcherQueue` to `JoinableTaskFactory`?

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/WinUI3TemplateCs (Package).wapproj
```diff
@@ -0,0 +1,70 @@
...
+      <Platform>arm64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup>
+    <WapProjPath Condition="'$(WapProjPath)'==''">$(MSBuildExtensionsPath)\Microsoft\DesktopBridge\</WapProjPath>
+    <PathToXAMLWinRTImplementations>WinUI3TemplateCs\</PathToXAMLWinRTImplementations>
+  </PropertyGroup>
+  <Import Project="$(WapProjPath)\Microsoft.DesktopBridge.props" />
+  <PropertyGroup>
+    <ProjectGuid>17b55766-7f58-4604-ac56-faa983ee25cb</ProjectGuid>
```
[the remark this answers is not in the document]
angelazhangmsft: Seems like we can remove this.

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/WinUI3TemplateCs (Package).wapproj
```diff
@@ -0,0 +1,70 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\WinUI3TemplateCs\WinUI3TemplateCs.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <Content Include="Images\logo.png" />
+    <Content Include="Images\SplashScreen.png" />
+    <Content Include="Images\Square150x150Logo.png" />
+    <Content Include="Images\Square44x44Logo.png" />
+    <Content Include="Images\Wide310x150Logo.png" />
```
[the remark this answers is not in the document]
angelazhangmsft: good suggestion :)

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/WinUI3TemplateCs (Package).wapproj
```diff
@@ -0,0 +1,70 @@
...
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|arm64">
+      <Configuration>Debug</Configuration>
+      <Platform>arm64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|arm64">
+      <Configuration>Release</Configuration>
+      <Platform>arm64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
```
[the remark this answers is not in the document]
AArnott: Who owns the wapproj project type? Particularly if it's new for Maui, I'd very much like to meet to discuss all the old patterns this project type is using and see if we can modernize it before shipping.

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/WinUI3TemplateCs (Package).wapproj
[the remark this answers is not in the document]
AArnott: Another sign of an older project pattern. I'd like to explore this a bit more with the owner of the project type.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml.cpp
angelazhangmsft: @jonwis, do you recommend using the same `.editorconfig` from the WinAppSDK repo?
https://github.com/microsoft/WindowsAppSDK/blob/main/.editorconfig

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/WinUI3TemplateCs (Package).wapproj
```diff
@@ -0,0 +1,70 @@
...
+    <AppxManifest Include="Package.appxmanifest">
+      <SubType>Designer</SubType>
+    </AppxManifest>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ProjectReunion" Version="[0.8.2]">
+      <IncludeAssets>build</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.ProjectReunion.WinUI" Version="[0.8.2]">
+      <IncludeAssets>build</IncludeAssets>
```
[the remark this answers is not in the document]
angelazhangmsft: From offline - there’s no active requirement for including only build assets – it’s simply the only thing needed (props & targets). As such, we can leave it as is

SharedContent/cs-winui-template/WinUI3TemplateCs/MainWindow.xaml.cs
angelazhangmsft: This results in the error: 'MONITORINFO' does not have a predefined size, therefore sizeof can only be used in an unsafe context

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
angelazhangmsft: May consider staying with assigning the property in the SampleConfiguration.cs file, since that is the file we would ask sample authors to edit.

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
angelazhangmsft: Since all samples will have at least 1 scenario, changing from `FirstOrDefault` to `First`

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
angelazhangmsft: Looks like `as<>`  doesn't exist

SharedContent/cpp-winui-template/WinUI3TemplateCpp/App.xaml
```diff
@@ -0,0 +1,134 @@
...
+    xmlns:local="using:WinUI3TemplateCpp">
+    <Application.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
+                <!-- Other merged dictionaries here -->
+            </ResourceDictionary.MergedDictionaries>
+            <!-- Other app resources here -->
+
+            <Style x:Key="SymbolButton" TargetType="ToggleButton">
```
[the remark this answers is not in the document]
angelazhangmsft: Factored these out

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainPage.xaml.cpp
```diff
@@ -0,0 +1,155 @@
...
+#if __has_include("MainPage.g.cpp")
+#include "MainPage.g.cpp"
+#endif
+
+using namespace winrt;
+using namespace Microsoft::UI::Xaml;
+using namespace Windows::UI::Xaml::Interop;
+using namespace Microsoft::UI::Xaml::Media;
+using namespace Microsoft::UI::Xaml::Media::Animation;
+using namespace Microsoft::UI::Xaml::Controls;
```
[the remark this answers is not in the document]
angelazhangmsft: Alphabetized, with exception of `using winrt;` which is always first

SharedContent/cs-winui-template/WinUI3TemplateCs/WinUI3TemplateCs.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <RootNamespace>WinUI3TemplateCs</RootNamespace>
```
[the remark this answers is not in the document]
angelazhangmsft: Dropped `RootNamespace`. Not sure if this will change depending on how the templates are distributed

SharedContent/cs-winui-template/WinUI3TemplateCs/SettingsPage.xaml.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.UI.Xaml.Navigation;
+
+using System;
+using System.Linq;
+using System.Reflection;
+
+
+namespace WinUI3TemplateCs
+{
+    public sealed partial class SettingsPage : Page
```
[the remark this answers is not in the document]
angelazhangmsft: removed all instances of `sealed`

SharedContent/cs-winui-template/WinUI3TemplateCs/SettingsPage.xaml
```diff
@@ -0,0 +1,41 @@
...
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <Grid RowDefinitions="*,Auto">
+
+        <StackPanel HorizontalAlignment="Left" Spacing="20" >
+            <TextBlock  Style="{StaticResource TitleTextBlockStyle}"
+                                Margin="0,4,0,0"
```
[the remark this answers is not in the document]
angelazhangmsft: fixed

SharedContent/cs-winui-template/WinUI3TemplateCs/MainWindow.xaml.cs
```diff
@@ -0,0 +1,110 @@
...
+            const int ICON_BIG = 1;
+
+            fixed (char* nameLocal = iconName)
+            {
+                HINSTANCE hInstLocal = default(HINSTANCE);
+                HANDLE __result = LoadImage(hInstLocal, nameLocal,
+                    GDI_IMAGE_TYPE.IMAGE_ICON,
+                    GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CXSMICON),
+                    GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CYSMICON),
+                    IMAGE_FLAGS.LR_LOADFROMFILE);
```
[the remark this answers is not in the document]
angelazhangmsft: added

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
angelazhangmsft: Changed to `page` and removed initialization

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
angelazhangmsft: changed from `auto` to `UINT`

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml
angelazhangmsft: According to https://docs.microsoft.com/windows/apps/design/controls/radio-button#create-your-own-radiobutton-group, it is implicitly grouped since they are in the same stack panel

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/WinUI3TemplateCs (Package).wapproj
```diff
@@ -0,0 +1,70 @@
...
+    <AppxPackageSigningEnabled>True</AppxPackageSigningEnabled>
+    <EntryPointProjectUniqueName>..\WinUI3TemplateCs\WinUI3TemplateCs.csproj</EntryPointProjectUniqueName>
+  </PropertyGroup>
+  <ItemGroup>
+    <AppxManifest Include="Package.appxmanifest">
+      <SubType>Designer</SubType>
+    </AppxManifest>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ProjectReunion" Version="[0.8.2]">
```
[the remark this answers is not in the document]
angelazhangmsft: I think we can remove the brackets - note that these templates will need to upgrade to 1.0 (Microsoft.WindowsAppSDK) as well

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
angelazhangmsft: fixed by QI directly with IWindowNative

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml.cpp
angelazhangmsft: Changed to use `ElementTheme` instead of strings.

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml.cpp
angelazhangmsft: added `.editorconfig` from the WinAppSDK repo.

SharedContent/cs-winui-template/WinUI3TemplateCs/SampleConfiguration.cs
```diff
@@ -0,0 +1,38 @@
...
+//
+//*********************************************************
+
+using System;
+using System.Collections.Generic;
+using Microsoft.UI.Xaml.Controls;
+
+namespace WinUI3TemplateCs
+{
+    public static class Settings
```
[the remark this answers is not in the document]
angelazhangmsft: made this Settings class `internal`. Not sure if we want any other classes to be internal, most of the other classes are created using default page templates.

SharedContent/README.md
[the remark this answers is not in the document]
angelazhangmsft: We will work on creating a VSIX for the samples

SharedContent/cs-winui-template/WinUI3TemplateCs (Package)/Package.appxmanifest
```diff
@@ -0,0 +1,52 @@
...
+<Package
+  xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+  IgnorableNamespaces="uap rescap">
+
+  <Identity
+    Name="cfb926cc-c146-49f1-9ac1-cfaf8182bfd7"
+    Publisher="CN=Microsoft Corporation"
+    Version="1.0.0.0" />
```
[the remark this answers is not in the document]
angelazhangmsft: Will change the Name value when creating VSIX template. As for Publisher, I am modeling off of https://github.com/microsoft/Windows-universal-samples/blob/ca74baf5c9e53955e8d749733b8e3e4839afd472/SharedContent/Templates/UWPSDKSampleCS/Package.appxmanifest#L10

SharedContent/cpp-winui-template/WinUI3TemplateCpp/SettingsPage.xaml
angelazhangmsft: Added `.editorconfig` and `prettierc` files to the repo

SharedContent/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+    MainWindow::MainWindow()
+    {
+        InitializeComponent();
+
+        Title(winrt::WinUI3TemplateCpp::Settings::FeatureName);
+
+        HWND hwnd = GetWindowHandle();
+
+        LoadIcon(hwnd, L"windows-sdk.ico");
```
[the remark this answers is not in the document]
angelazhangmsft: Recommendation for WinAppSDK is calling the MUI APIs: https://github.com/microsoft/microsoft-ui-xaml/issues/4056#issuecomment-774415870

SharedContent/cs-winui-template/WinUI3TemplateCs/MainPage.xaml.cs
angelazhangmsft: From offline thread we will continue using DQ

Templates/VSIX/Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <CppWinRTVersion Condition="'$(CppWinRTVersion)' == ''">2.0.210922.5</CppWinRTVersion>
+    <CsWin32Version Condition="'$(CsWin32Version)' == ''">0.1.506-beta</CsWin32Version>
+    <WindowsSDKBuildToolsVersion Condition="'$(WindowsSDKBuildToolsVersion)' == ''">10.0.20348.19</WindowsSDKBuildToolsVersion>
```
evelynwu-msft, line 7: >10.0.20348.19 [](http://example.com/codeflow?start=84&length=13)

`10.0.22000.194` was just published, so you'll want to use that instead.

Templates/VSIX/WindowsAppSDKSampleVSIX.csproj
```diff
@@ -0,0 +1,106 @@
...
+    </Content>
+    <None Include="source.extension.vsixmanifest" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Windows.CppWinRT" Version="[$(CppWinRTVersion)]" GeneratePathProperty="true">
+      <ExcludeAssets>All</ExcludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.Windows.CsWin32" Version="[$(CsWin32Version)]" GeneratePathProperty="true">
+      <ExcludeAssets>All</ExcludeAssets>
+    </PackageReference>
```
evelynwu-msft, line 66: Is this needed?

Templates/VSIX/Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <CppWinRTVersion Condition="'$(CppWinRTVersion)' == ''">2.0.210922.5</CppWinRTVersion>
+    <CsWin32Version Condition="'$(CsWin32Version)' == ''">0.1.588-beta</CsWin32Version>
```
evelynwu-msft, line 6: ><CsWin32Version Condition="'$(CsWin32Version)' == ''">0.1.588-beta</CsWin32Version> [](http://example.com/codeflow?start=4&length=83)

Is this needed?

Templates/VSIX/WindowsAppSDKSampleVSIX.csproj
angelazhangmsft: removed

Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.h
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#pragma once
+#include "pch.h"
+
+namespace winrt::$safeprojectname$
+{
+    class Settings
```
jonwis, line 9: Consider: use `struct` instead of `class` and call this `SampleConfig` maybe.

Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/Scenario1_ShortName.xaml.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    MainPage Scenario1_ShortName::rootPage{ nullptr };
+
+    Scenario1_ShortName::Scenario1_ShortName()
+    {
+        InitializeComponent();
+        Scenario1_ShortName::rootPage = MainPage::Current();
+    }
+
+    void Scenario1_ShortName::NotifyUser(hstring const& strMessage, InfoBarSeverity const& severity)
```
jonwis, line 27: In the Windows samples, the "notify user" machinery goes on rootPage - so you say `rootPage.NotifyUser(...)` as the info bar belongs to the "containing" page of the sample - https://github.com/microsoft/Windows-universal-samples/blob/29d29659ac2821e5718024208ac1411535ff0785/Samples/AppServices/cppwinrt/AppServicesClient/KeepConnectionOpenScenario.cpp#L47-L50 has an example of this.

Templates/VSIX/Project Templates/cs-winui-template/Properties/PublishProfiles/win10-x64.pubxml
```diff
@@ -0,0 +1,19 @@
...
+<!--
+https://go.microsoft.com/fwlink/?LinkID=208121.
+-->
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <PublishProtocol>FileSystem</PublishProtocol>
+    <Platform>x64</Platform>
+    <RuntimeIdentifier>win10-x64</RuntimeIdentifier>
+    <PublishDir>bin\$(Configuration)\$(TargetFramework)\$(RuntimeIdentifier)\publish\</PublishDir>
+    <SelfContained>true</SelfContained>
```
jonwis, line 11: Do the samples produce "all four" styles of sample? Packaged, Unpackaged, Evergreen, BYO? Or just Packaged + Evergreen?

Templates/VSIX/WindowsAppSDKSampleVSIX.sln
```diff
@@ -0,0 +1,77 @@
...
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "WinUI3ScenarioPageCpp", "Item Templates\WinUI3ScenarioPageCpp\WinUI3ScenarioPageCpp.csproj", "{538FC0F0-5250-453E-B40F-40881A965490}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "WinUI3ScenarioPageCs", "Item Templates\WinUI3ScenarioPageCs\WinUI3ScenarioPageCs.csproj", "{C061FD19-1AF2-4FAE-BD42-A9BC8129B464}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Debug|x86 = Debug|x86
+		Release|Any CPU = Release|Any CPU
```
jonwis, line 26: Does this support ARM64?

jonwis (approved):

Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/MainPage.xaml.h
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#pragma once
+#include "MainPage.g.h"
+
+namespace winrt::$safeprojectname$::implementation
```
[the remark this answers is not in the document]
angelazhangmsft: According to the docs [here](https://docs.microsoft.com/en-us/visualstudio/ide/template-parameters?view=vs-2022), $rootnamespace$ only applies to item templates and not project templates

Templates/VSIX/WindowsAppSDKSampleVSIX.sln
angelazhangmsft: The sample templates themselves include ARM64, but looks like the VSIX project itself doesn't need to.

For example here is the VSIX solution for WinAppSDK templates: https://github.com/microsoft/WindowsAppSDK/blob/main/dev/VSIX/WindowsAppSDK.Extension.sln

@evelynwu-msft do you know why we don' need ARM64 here?

jonwis (approved):

PR #75, opened 2021-09-10
Create .gitattributes

angelazhangmsft:


btueffers (approved):

erik0686:
@angelazhangmsft  Maybe I'm totally wrong about this, but should the files that have possibly been changed to crlf be pushed again? I'm seeing changes from my local files to the ones in the repo, and I think it could possibly be because of this change :) 

Please let me know if what I'm saying is total rubbish though :)

erik0686:
For context, right after cloning the repo, I see git marking several files as modified. For example:
https://github.com/microsoft/WindowsAppSDK-Samples/compare/user/ertorres/Samples?expand=1

Although after thinking about it, maybe it could be my git settings? Please let me know if you have any ideas :)

PR #90, opened 2021-09-29
Instead of crashing, print an error message if running unpackaged console app elevated.

rohanp-msft:
### How verified

**Before**
image

**Now:**
```
PS S:\Project-Reunion-Samples> S:\Project-Reunion-Samples\Samples\ResourceManagement\cpp-console-unpackaged\x64\Debug\console_unpackaged_app.exe default
Windows App SDK doesn't currently support elevation for unpackaged apps. See https://github.com/microsoft/WindowsAppSDK/issues/567.
PS S:\Project-Reunion-Samples>
```

App still works in unelevated mode:
```
PS C:\Users\rohanp> S:\Project-Reunion-Samples\Samples\ResourceManagement\cpp-console-unpackaged\x64\Debug\console_unpackaged_app.exe Override
Deutsches Beispiel
PS C:\Users\rohanp>
```

### DIFF
--- a/Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.cpp
+++ b/Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.cpp
@@ -1,7 +1,8 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include <windows.h>
+#include <shlobj_core.h>
 #include <wil/resource.h>
 #include <iostream>
 
@@ -56,6 +57,13 @@ int wmain(int argc, wchar_t* argv[])
             "    console_unpackaged_app.exe override\n"
             "  Get the sample string for the resource-not-found fallback case\n"
             "    console_unpackaged_app.exe fallback\n";
+        return 1;
+    }
+
+    // Despite the name of this API, it will return FALSE if the user is an admin but the context isn't elevated.
+    if (IsUserAnAdmin())
+    {
+        std::wcout << "Windows App SDK doesn't currently support elevation for unpackaged apps. See https://github.com/microsoft/WindowsAppSDK/issues/567." << std::endl;
         return 1;
     }
 


huichen123:
>     HRESULT hr{ MddBootstrapInitialize(majorMinorVersion, versionTag, minVersion) };

@DrusTheAxe can MddBootstrapInitialize return a failure if elevation is not supported?

---
Refers to: Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.cpp:25 in e4b1bfe. [](commit_id = e4b1bfe5f573f807a63f517929c60affa9b06ecc, deletion_comment = False)

rohanp-msft:
> > ```
> > HRESULT hr{ MddBootstrapInitialize(majorMinorVersion, versionTag, minVersion) };
> > ```
> 
> @DrusTheAxe can MddBootstrapInitialize return a failure if elevation is not supported?
> 
> Refers to: Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.cpp:25 in [e4b1bfe](https://github.com/microsoft/WindowsAppSDK-Samples/commit/e4b1bfe5f573f807a63f517929c60affa9b06ecc). [](commit_id = [e4b1bfe](https://github.com/microsoft/WindowsAppSDK-Samples/commit/e4b1bfe5f573f807a63f517929c60affa9b06ecc), deletion_comment = False)

@huichen123 I agree that the right solution here is for DD to return an error. At this point though, I don't think it meets the bar for v1.0 Preview 3 or stable. When stable v1.0 is released, it would be good for the samples to be in good shape. So I think we should take this change. I can add an issue in this repo to undo this once DD supports elevated or modify this change once DD returns an error. What do you think?
CC @DrusTheAxe

Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.cpp
```diff
@@ -59,6 +60,13 @@ int wmain(int argc, wchar_t* argv[])
         return 1;
     }
 
+    // Despite the name of this API, it will return FALSE if the user is an admin but the context isn't elevated.
+    if (IsUserAnAdmin())
```
DrusTheAxe, line 64: What's this? Where's the implementation?

What does "is an admin" mean? Can only such admins be elevated?
See FailFastIfElevated for the actual check done by DynDep that you're trying to work around

Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.cpp
```diff
@@ -59,6 +60,13 @@ int wmain(int argc, wchar_t* argv[])
...
+    {
+        std::wcout << "Windows App SDK doesn't currently support elevation for unpackaged apps. See https://github.com/microsoft/WindowsAppSDK/issues/567." << std::endl;
```
DrusTheAxe, line 66: `for unpackaged apps`

Dynamic Dependencies is the issue, thus unpackaged apps using WinAppSDK-as-MSIX. If/when SelfContained unpackaged apps are supported this wouldn't be relevant (purpose in life is to give an unpackaged app access to packaged goods. No WinAppSDK/MSIX package then no need for WinAppSDK's DynDep).

SUGGESTION: "Windows app SDK DynamicDependencies doesn't support elevation. See Issue #567 https://github.com/microsoft/WindowsAppSDK/issues/567"

Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.cpp
DrusTheAxe, line 64: I don't recommend doing this. Line 24 calls the bootstrapper. Today that does a FAIL_FAST_IF(is_elevated) so you'll never get here.

I'm changing MddBootstrapInitialize to return `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)` instead of FailFast so you could trap for that error explicitly, or just do nothing and let the current error handling handle it.

I recommend that rather than doing any explicitly elevation check like you're doing here

DrusTheAxe (changes requested):

rohanp-msft:
Elevation support is coming in 1.1, so this crashing behavior will be gone then. So, closing this PR.

PR #91, opened 2021-10-01
Adding files that appear modified after a git pull

ZenBird:


rohanp-msft:
Does this work? Can you add a description of the verification to the PR description, please?
Also, for those like me that already did a git pull, what's the way forward? I mean, I can't make the files go away. `git restore .` doesn't remove them. Without removing them, I can't do a `git pull` to pick up your fix once it's checked in.

ZenBird:
> Thanks for the quick response! If possible, it would be good to verify before check-in if this fixes the issue. Otherwise I'm OK with checking this in to see if it fixes it.

I am not able to verify it since it does not repro in my local branch (because the files are committed) and we cannot verify on main until the files are merged. It looks like the .editorconfig file in WindowsAppSDK-Samples may be causing this issue. If that is the case, then this will resolve it since these files are formatted according to the new .editorconfig. I do not know how the .editorconfig works. I will research this before confirming if this will work.

rohanp-msft (approved):

PR #92, opened 2021-10-01
Enable use of different build tool versions depending on VS version

erik0686:
For our task of ensuring the Windows App SDK samples build correctly in preview versions of VS (currently, VS 2022 preview), we needed to add support to use the corresponding version of BuildTools (v143 for VS2022) while keeping the support for previous versions.
@jefgen  Had a cool idea of detecting the versions of VS being used to build the samples and use the corresponding versions of the build tools, without the need of having them all installed at the same time. For this, I added a props file that will set the <PlatformToolSet> variable to whatever it needs to be. 

In this PR, I also added a nuget.config file, since to restore packages in the pipeline build, it was needed to get the package sources. 

To verify that using different versions of the buildtools was working, I added (for debugging and then removed) a message that would show up when building. 
```  
<Target Name="Display platform toolset" AfterTargets="Build">
      <Message Importance="high" Text="for platform $(BuildToolVersion), use toolset: $(PlatformToolset)" />
  </Target>
```
When building in VS2019, this would show up as:
          "for platform 16.0, use toolset: v142"
When building with VS2022:
          "for platform 17.0, use toolset: v143"


jefgen:
Thanks for doing this Erik! :)

Samples/Build.Windows.PlatformToolset.props
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (C) 2016 and later: Unicode, Inc. and others. License & terms of use: http://www.unicode.org/copyright.html -->
```
erik0686, line 2: Sigh... the problems of copying and pasting... I'll fix this

jefgen, line 2: Hm, this should probably be the MIT License:
```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License. See LICENSE in the project root for license information.
```

Samples/Build.Windows.PlatformToolset.props
```diff
@@ -0,0 +1,29 @@
...
+  <!-- This file is used to set configuration option for all projects. -->
+
+  <!-- Automatic PlatformToolset version selection. -->
+  <!-- If there is no DefaultPlatformToolset set, we will try to detect the version based on version of the build tools. -->
+  <PropertyGroup>
+    <BuildToolVersion>$(VisualStudioVersion)</BuildToolVersion>
+    <!-- Only use the MSBuildToolsVersion if we don't have the VisualStudioVersion and MSBuildToolsVersion is set to something other than Current. -->
+    <BuildToolVersion Condition="'$(BuildToolVersion)'=='' and '$(MSBuildToolsVersion)' != 'Current'">$(MSBuildToolsVersion)</BuildToolVersion>
+    <!-- Note:
+      v140 is the Visual Studio 2015 toolset. (14.0)
```
jefgen, line 13: Does WinAppSDK build/support VS 2015 or VS 2017?
I wonder if we should perhaps only have VS2019 and VS2022.

erik0686: Very good question, I could definitely try, but you're probably right

rohanp-msft: I think we don't support those versions based on [https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable#2-install-visual-studio](https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable#2-install-visual-studio). I think we should remove them.

Samples/Windowing/cs-winui/SampleApp (Package)/SampleApp (Package).wapproj
```diff
@@ -42,7 +42,7 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <AppxPackageSigningEnabled>True</AppxPackageSigningEnabled>
     <EntryPointProjectUniqueName>..\SampleApp\SampleApp.csproj</EntryPointProjectUniqueName>
-    <PackageCertificateThumbprint>52902158A7BB4EF57D9DB04AF265DB38A11F1661</PackageCertificateThumbprint>
+    <PackageCertificateThumbprint>697B0A4425F6E61C280403E9B3F468E3E3C905A5</PackageCertificateThumbprint>
```
rohanp-msft, line 45: Is this change intentional, Erik?

Samples/Build.Windows.PlatformToolset.props
angelazhangmsft: WinAppSDK only supports VS2019 and VS 2022, so I think it would make sense to only have those

Samples/Windowing/cs-winui/SampleApp (Package)/SampleApp (Package).wapproj
erik0686: Yes, intentional, although maybe optional? When building, VS was giving warnings about the certificate being expired. So I regenerated it. 
They were warnings though, not errors, so if you think I should undo this change, I will.

Samples/Build.Windows.PlatformToolset.props
erik0686: Done! :) Thanks!!

Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.vcxproj
```diff
@@ -3,6 +3,7 @@
   <Import Project="packages\Microsoft.WindowsAppSDK.1.0.0-preview1\build\native\Microsoft.WindowsAppSDK.props" Condition="Exists('packages\Microsoft.WindowsAppSDK.1.0.0-preview1\build\native\Microsoft.WindowsAppSDK.props')" />
   <Import Project="packages\Microsoft.Windows.SDK.BuildTools.10.0.20348.19\build\Microsoft.Windows.SDK.BuildTools.props" Condition="Exists('packages\Microsoft.Windows.SDK.BuildTools.10.0.20348.19\build\Microsoft.Windows.SDK.BuildTools.props')" />
   <Import Project="packages\Microsoft.Windows.CppWinRT.2.0.210714.1\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('packages\Microsoft.Windows.CppWinRT.2.0.210714.1\build\native\Microsoft.Windows.CppWinRT.props')" />
+  <Import Project="..\..\Build.Windows.PlatformToolset.props" />
```
jefgen, line 6: I think this might need to be 
`  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />` to work in some cases.

(for all project files... sorry for the churn...)

Samples/Windowing/cs-winui/SampleApp (Package)/SampleApp (Package).wapproj
angelazhangmsft: I wonder if we should remove this and set `AppxPackageSigningEnabled` to False like in this sample: https://github.com/microsoft/WindowsAppSDK-Samples/blob/c8070a0d1710d19f6e870f2d3ec12e5422b3da90/Samples/ResourceManagement/cs-winui/winui_desktop_packaged_app%20(Package)/winui_desktop_packaged_app%20(Package).wapproj#L43

Although that is a separate issue.

erik0686: I actually do think that is a good idea. Having it be true, means that in the pipeline build it will fail, so when running the pipeline I have to do something like:
```MsbuildArgs: '/p:VCToolsInstallDir="$(VCToolsInstallDir)\" /p:AppxPackageSigningEnabled=false'```
If we can avoid it in the samples themselves, that would reduce a little the churn we have... 
what do you think?

File: Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.vcxproj (hunk 4 above)
Comment (erik0686) [reply Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.vcxproj:6]: Hmmm I'm not sure I understood...

Do you mean that whenever we import Build.Windows.PlatformToolset.props, we should add the line you mentioned right before it?

Comment (jefgen) [reply Samples/ResourceManagement/cpp-console-unpackaged/console_unpackaged_app.vcxproj:6]: Ah, sorry for not being clear...

What I meant was that this:
`<Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />`
must come *before* this:
`<Import Project="..\..\Build.Windows.PlatformToolset.props" />`

However, doing some more investigating, I'm not sure if this is really needed or not though.
If the samples work fine, then it's probably not needed.

File: Samples/Windowing/cs-winui/SampleApp (Package)/SampleApp (Package).wapproj (hunk 3 above)
Comment (angelazhangmsft) [reply Samples/Windowing/cs-winui/SampleApp (Package)/SampleApp (Package).wapproj:45]: I agree that we should turn off signing and follow the example from the existing ResourceManagement samples

File: Samples/Build.Windows.PlatformToolset.props [hunk 5]
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rohanp-msft, line 1: I think the name of this file can be improved:
- We *may* want to make this file general-purpose. That way, we can add more helper MSBuild "code" like this in the future if needed.
- It's C++ specific, it'd help clarity-wise to include that in the name. Not everyone looking at the file will know it's C++ specific based on `PlatformToolset`.
- I don't know that it makes sense to say Windows in the name. All our samples are for Windows. Even if this file was `PlatformToolset` specific: `PlatformToolset` doesn't apply to Windows in general, only to UWPs.

jefgen: +1 good suggestions.
I think the name was just copied from my other implementation, so probably that where it originally came from.

jefgen: > doesn't apply to Windows in general, only to UWPs.

Wouldn't we want to set the `PlatformToolset` for all application samples, regardless of UWP or not?

rohanp-msft: I don't know much about this toolset. It's based on what I see in the VS installer that I said that it's UWP specific:
image
That does then beg the question of whether it's needed in these non-UWPs... @erik0686, can I leave it to you to learn more about this toolset, whether we're needlessly including it in these apps, etc.?

jefgen: I think the "v142" or "v143" for the `PlatformToolset` just refers to the version of the [MSVC compiler/linker/etc.](https://docs.microsoft.com/en-us/cpp/build/how-to-modify-the-target-framework-and-platform-toolset?view=msvc-160#:~:text=The%20platform%20toolset%20consists%20of%20the%20C%2B%2B%20compiler,linker%20%28link.exe%29%2C%20along%20with%20the%20C%2FC%2B%2B%20standard%20libraries.) that get used to build... So it could be used for either UWP or not. I'm not sure why the VS Installer makes it look like they are UWP-only...  🤔

FWIW, I've previously used the same auto-detection logic for "Classic Desktop C++" projects and it worked fine. :)

erik0686: What would you think of a generic name such as Build.Helpers.PlatformToolset.props ??? would that sound as a better name?

rohanp-msft: I looked in "Desktop Development with C++" and I see this:
image
I don't know if "MSVC v142" is a separate toolset or not...
Assuming they're not, I also don't know which toolset is referenced when you include `<PlatformToolset>v142</PlatformToolset>` in your project file.

rohanp-msft: > Assuming they're not, I also don't know which toolset is referenced when you include <PlatformToolset>v142</PlatformToolset> in your project file.

Correcting something I said. If there are two different toolsets, presumably the right one is selected based on whether the app is a UWP or not.

I think PlatformToolset is automatically added to the project file. So it must be needed.

So I think we're all good here. Regarding the name, Erik, about: Build.Common.Cpp.props?

rohanp-msft: @erik0686 Given all this, it seems to me that the proper way to verify the change would be to: build in VS cmd and look in the binlog to see which compiler was used!

jefgen: FWIW, I like Rohan's suggestion of `Build.Common.Cpp.props` for the file name.

erik0686: Thanks for the suggestions, @rohanp-msft,
For the PlatformToolset, I can see the version is correctly being selected, in the binlog file, I can see this:
image

As for the MSVC Version being used, I'm not exactly sure if this is what you were asking for, but I see this:
image
Does this correctly answer your question? I can recheck for the correct things if they were not.

rohanp-msft: That's it, @erik0686. Thanks. Here's why I think that's right:

I see "The toolset version is v140 for Visual Studio 2015, v141 for 2017, and v142 for 2019." in https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-160.

On my machine, I see:
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133
C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.30.30423

So I think v\d\d\d is based on the first three digits of the full version number.

So I think 14.29.30133 being used for your selection v142  is right.

Relevant fact that I learned from the link above: I learned that the major number (14) being the same after VS 2013 is to indicate binary compatibility of binaries built using different version 14.* toolsets.

erik0686: Ohhhh that is really cool info, Rohan! thanks for sharing :)

jefgen (approved):
Thanks for doing this Erik! :)

docs/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
+<!--
+Thank you for your pull request!
+
+Please see https://github.com/microsoft/WindowsAppSDK-Samples/blob/main/docs/samples-guidelines.md for guidelines on
+how to best contribute to the Windows App SDK Samples!
```
rohanp-msft, line 5: >to the Windows App SDK Samples [](http://example.com/codeflow?start=23&length=30)

Grammar-wise, I think we should use "to the Windows App SDK Samples project" or "to Windows App SDK Samples". I think the former is better (since it's explicit).

docs/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
...
+
+-->
+
+##### Checklist
+
+- [ ] Required: Please ensure youre samples can be correctly built using Visual Studio 2019 and above
```
rohanp-msft, line 11: >youre [](http://example.com/codeflow?start=30&length=5)

your

rohanp-msft, line 11: >Visual Studio 2019 and above [](http://example.com/codeflow?start=73&length=28)

maybe this instead? that way we don't have to come back here and update this later when we change the min version?
"the Visual Studio versions in [https://docs.microsoft.com/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable#2-install-visual-studio](https://docs.microsoft.com/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable#2-install-visual-studio)"

docs/pull_request_template.md
angelazhangmsft: Windows App SDK Samples repository?

Samples/Build.Common.Cpp.props
```diff
@@ -0,0 +1,26 @@
...
+
+  <!-- Automatic PlatformToolset version selection. -->
+  <!-- If there is no DefaultPlatformToolset set, we will try to detect the version based on version of the build tools. -->
+  <PropertyGroup>
+    <BuildToolVersion>$(VisualStudioVersion)</BuildToolVersion>
+    <!-- Only use the MSBuildToolsVersion if we don't have the VisualStudioVersion and MSBuildToolsVersion is set to something other than Current. -->
+    <BuildToolVersion Condition="'$(BuildToolVersion)'=='' and '$(MSBuildToolsVersion)' != 'Current'">$(MSBuildToolsVersion)</BuildToolVersion>
+    <!-- Note:
+      v142 is the Visual Studio 2019 toolset. (16.0)
+      v143 is the Visual Studio 2019 toolset. (17.0)
```
rohanp-msft, line 15: >2019 [](http://example.com/codeflow?start=32&length=4)

2022

Samples/Build.Common.Cpp.props
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- This file is used to set configuration option for all projects. -->
```
rohanp-msft, line 5: >for all projects. [](http://example.com/codeflow?start=53&length=17)

This part doesn't make sense to me, TBH. The file will set it wherever it is imported, right?

docs/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
...
+- [ ] If your sample is related to one of the existing scenarios (ResourceManagement, Windowing, etc)
```
angelazhangmsft, line 12: Not sure if we need this checklist point, since it's covered in docs/guidelines.md?

Samples/nuget.config
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <packageRestore>
+        <!-- Allow NuGet to download missing packages -->
+        <add key="enabled" value="True" />
+        <!-- Automatically check for missing packages during build in Visual Studio -->
```
rohanp-msft, line 6: It may be good to add that this file is added for the CI pipeline. I'm not 100%, but I think, by default, these properties are true when building locally in VS.

docs/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
+<!--
+Thank you for your pull request!
+
+Please see https://github.com/microsoft/WindowsAppSDK-Samples/blob/main/docs/samples-guidelines.md for guidelines on
+how to best contribute to the Windows App SDK Samples!
+
+-->
```
angelazhangmsft, line 7: Suggestion to also add a "Description" section here

Samples/nuget.config
```diff
@@ -0,0 +1,17 @@
...
+        <add key="automatic" value="True" />
+    </packageRestore>
+    <packageSources>
+    <clear />
+        <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    </packageSources>
+    <disabledPackageSources />
+    <activePackageSource>
+        <add key="All" value="(Aggregate source)" />
```
rohanp-msft, line 15: ><add key="All" value="(Aggregate source)" /> [](http://example.com/codeflow?start=8&length=44)

Curious, what does this do? Locally, I've never used it.

Samples/Build.Common.Cpp.props
```diff
@@ -0,0 +1,26 @@
...
+
+  <!-- Automatic PlatformToolset version selection. -->
+  <!-- If there is no DefaultPlatformToolset set, we will try to detect the version based on version of the build tools. -->
```
rohanp-msft, line 8: >the build tools [](http://example.com/codeflow?start=104&length=15)

VS/MSBuild tools? Just saying build tools is a bit confusing because `PlatformToolset` refers to MSVC build tools.

docs/pull_request_template.md
erik0686: Maybe I can make it clearer that this point is if the sample(s) are new :)

Samples/nuget.config
erik0686: This is actually not used :) I'll remove
In case you're curious
https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#package-source-sections

rohanp-msft (approved):
:shipit:

PR #95, opened 2021-10-05
Update Insights cpp-win32 app for different Build Tools versions

erik0686:
##### Description
When updating the samples to automatically detect which build tool version to use in my 
 previous PR (https://github.com/microsoft/WindowsAppSDK-Samples/pull/92) 
I failed to check if there had been any new samples added to the repo while my PR was still open. 
I noticed there's a new Sample that is not yet supporting this automatic detection, so I'm updating it :) 
This sample was added before I updated the PR Template to remind developers about this, so it was an unfortunate coincidence :) 
With this change though, the app now builds in both VS2019 and VS2022 using Build tools v142 and v143 respectively. 

##### Checklist

- [X] Please ensure your samples can be correctly built using the Visual Studio versions
      in the Visual Studio versions in https://docs.microsoft.com/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable#2-install-visual-studio
- [] If you're adding a new sample, and it is related to one of the existing scenarios 
- [X] Samples should build on all supported platforms (x64, x86, ARM64) and configurations (Debug, Release). 
- [X] Samples should set the minimum version to Windows 10 version 1809.
- [X] Make sure samples build clean with no warnings or errors.

### DIFF
--- a/Samples/Insights/cpp-win32/InsightsSample.vcxproj
+++ b/Samples/Insights/cpp-win32/InsightsSample.vcxproj
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="packages\Microsoft.WindowsAppSDK.1.0.0-preview1\build\native\Microsoft.WindowsAppSDK.props" Condition="Exists('packages\Microsoft.WindowsAppSDK.1.0.0-preview1\build\native\Microsoft.WindowsAppSDK.props')" />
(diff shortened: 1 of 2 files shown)


ZenBird (approved):
:shipit:

rohanp-msft (approved):
Thanks, Erik. 
Along with this change, can you fix the bolded part in pull_request_template.md?
"Please ensure your samples can be correctly built **using the Visual Studio versions
in the Visual Studio versions** in https://docs.microsoft.com/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable#2-install-visual-studio"

PR #100, opened 2021-10-12
Origin/user/rkarman/appwindow preview2

rkarman:
##### Description
This PR updates the Windowing sample to use the Preview 2 package.
As part of this SDK version changes were made to the API surface, the C++ and C# samples have been updated to match the new API surface.
All sample files now also have a copyright header as per guidelines.
C++ sample was cleaned up to remove the build warnings where possible.

##### Future improvements/considerations for next update (not part of this PR)
The change of the API surface to no longer have an AppWindowConfiguration class means we should rework the samples to no longer have a "Config" page, but rather have these options in the Presenter page and group them properly for OverlappedPresenter. For now I do a check for .Kind and disable all switches if any other Presenter is applied than an Overlapped.

##### Checklist

- [x] Please ensure your samples can be correctly built using the Visual Studio versions
- [x] Samples should build on all supported platforms (x64, x86, ARM64) and configurations (Debug, Release). 
- [x] Samples should set the minimum version to Windows 10 version 1809.
- [/] Make sure samples build clean with no warnings or errors.
Some warnings remain for the C++ sample, but these are coming from the SDK header files and cannot be fixed by the sample itself.


rkarman:
Dang, my github clone did not notify that I had an update in main after creating my branch. I will abandon this PR, rebase with the updates that Sam made, and then submit updates from there. Sorry for the noise... :)

PR #102, opened 2021-10-13
fix configuration mapping for sample

erik0686:
##### Description
Fixes configuration mapping. Before, the sample wouldn't build in Release x64 because the mapping was looking like this:
image

After this change, it looks like this:
image

I built the sample in Release and Debug for both x64 and x86 and it now builds :) I also removed a few "using" directives that were repeated and gave warnings in the sample building. 
##### Checklist

- [x] Please ensure your samples can be correctly built using the Visual Studio versions
- [ ] If you're adding a new sample, and it is related to one of the existing scenarios 
- [x] Samples should build on all supported platforms (x64, x86, ARM64) and configurations (Debug, Release). 
- [x] Samples should set the minimum version to Windows 10 version 1809.
- [x] Make sure samples build clean with no warnings or errors.


jefgen (approved):
Thanks for fixing this @erik0686!

PR #107, opened 2021-10-19
Cleanup sample templates

angelazhangmsft:
##### Description
This PR cleans up some of the sample template code (mainly in the C++ templates).

This includes:
- refactoring `using namespace winrt`
- auto&& in range-based for loops to avoid extra copies
- using `auto` to avoid repeating `<T>`
- NULL -> nullptr 
- update HMONITOR functions.

##### Checklist

- [ ] Please ensure your samples can be correctly built using the Visual Studio versions
- [ ] If you're adding a new sample, and it is related to one of the existing scenarios 
- [ ] Samples should build on all supported platforms (x64, x86, ARM64) and configurations (Debug, Release). 
- [ ] Samples should set the minimum version to Windows 10 version 1809.
- [ ] Make sure samples build clean with no warnings or errors.

### DIFF
--- a/Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/App.xaml.cpp
+++ b/Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/App.xaml.cpp
@@ -6,32 +6,33 @@
 #include "App.xaml.h"
 #include "MainWindow.xaml.h"
 
-using namespace winrt;
-using namespace Microsoft::UI::Xaml;
-using namespace Microsoft::UI::Xaml::Controls;
-using namespace Microsoft::UI::Xaml::Navigation;
-using namespace Windows::Foundation;
-using namespace $safeprojectname$;
-using namespace $safeprojectname$::implementation;
-
-App::App()
+namespace winrt
 {
-    InitializeComponent();
-
-#if defined _DEBUG && !defined DISABLE_XAML_GENERATED_BREAK_ON_UNHANDLED_EXCEPTION
-    UnhandledException([](IInspectable const&, UnhandledExceptionEventArgs const& e)
-    {
-        if (IsDebuggerPresent())
-        {
-            auto errorMessage = e.Message();
-            __debugbreak();
-        }
-    });
-#endif
+    using namespace Windows::Foundation;
+    using namespace Microsoft::UI::Xaml;
 }
 
-void App::OnLaunched(LaunchActivatedEventArgs const&)
+namespace winrt::$safeprojectname$::implementation
 {
-    window = make<MainWindow>();
-    window.Activate();
+    App::App()
+    {
+        InitializeComponent();
+
+#if defined _DEBUG && !defined DISABLE_XAML_GENERATED_BREAK_ON_UNHANDLED_EXCEPTION
+        UnhandledException([](winrt::IInspectable const&, winrt::UnhandledExceptionEventArgs const& e)
+            {
+                if (IsDebuggerPresent())
+                {
+                    auto errorMessage = e.Message();
+                    __debugbreak();
+                }
+            });
+#endif
+    }
+
+    void App::OnLaunched(winrt::LaunchActivatedEventArgs const&)
+    {
+        window = winrt::make<MainWindow>();
+        window.Activate();
+    }
 }
--- a/Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
+++ b/Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
@@ -10,9 +10,6 @@
 #include "SampleConfiguration.h"
 #endif
 
-using namespace winrt;
-using namespace Microsoft::UI::Xaml;
-
 namespace winrt::$safeprojectname$::implementation
 {
     MainWindow::MainWindow()
@@ -22,7 +19,6 @@ namespace winrt::$safeprojectname$::implementation
         Title(winrt::$safeprojectname$::SampleConfig::FeatureName);
 
         HWND hwnd = GetWindowHandle();
-
         LoadIcon(hwnd, L"windows-sdk.ico");
         SetWindowSize(hwnd, 1050, 800);
         PlacementCenterWindowInMonitorWin32(hwnd);
@@ -32,7 +28,7 @@ namespace winrt::$safeprojectname$::implementation
     {
         if (_hwnd == nullptr)
         {
-            Window window = *this;
+            winrt::Microsoft::UI::Xaml::Window window = *this;
             window.as<IWindowNative>()->get_WindowHandle(&_hwnd);
         }
         return _hwnd;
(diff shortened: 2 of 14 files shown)


Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/App.xaml.cpp
```diff
@@ -6,32 +6,33 @@
...
 #include "MainWindow.xaml.h"
 
-using namespace winrt;
-using namespace Microsoft::UI::Xaml;
-using namespace Microsoft::UI::Xaml::Controls;
-using namespace Microsoft::UI::Xaml::Navigation;
-using namespace Windows::Foundation;
-using namespace $safeprojectname$;
-using namespace $safeprojectname$::implementation;
+namespace winrt
```
jonwis, line 9: @oldnewthing - the existing "slurk everything into the global namespace in a file" is the style in the OS samples. Out of curiosity, is there a meaningful difference?

Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp
```diff
@@ -40,63 +41,49 @@ namespace winrt::$safeprojectname$::implementation
...
+        const int h = (&prc)->bottom - (&prc)->top;
 
-        rc = mi.rcWork;
-        prc->left = rc.left + (rc.right - rc.left - w) / 2;
-        prc->top = rc.top + (rc.bottom - rc.top - h) / 2;
-        prc->right = prc->left + w;
-        prc->bottom = prc->top + h;
+        (&prc)->left = rcWork.left + (rcWork.right - rcWork.left - w) / 2;
+        (&prc)->top = rcWork.top + (rcWork.bottom - rcWork.top - h) / 2;
+        (&prc)->right = (&prc)->left + w;
```
jonwis, line 86: These are more naturally written as just `prc.right = prc.left + w;` - you don't need to take the address of the reference to use `operator->` on them.  Except for the call to `MonitorFromRect` which does need `RECT*` as the parameter type.

(Also, consider renaming `prc` to just `rc`, or `rect`)

Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.cpp
```diff
@@ -6,16 +6,20 @@
...
-	{
-		Scenario{ L"Scenario 1", winrt::hstring(winrt::name_of<$safeprojectname$::Scenario1_ShortName>())},
-		Scenario{ L"Scenario 2", winrt::hstring(winrt::name_of<$safeprojectname$::Scenario2_ShortName>())}
-	});
+namespace winrt::$safeprojectname$
+{
+    IVector<Scenario> implementation::MainPage::scenariosInner = winrt::single_threaded_observable_vector<Scenario>(
+        {
+            Scenario{ L"Scenario 1", winrt::hstring(winrt::name_of<$safeprojectname$::Scenario1_ShortName>())},
+            Scenario{ L"Scenario 2", winrt::hstring(winrt::name_of<$safeprojectname$::Scenario2_ShortName>())}
```
jonwis, line 20: Huh, I'm surprised you need this constructor here - what happens if you just say:

```c++
Scenario{ L"Scenario 1", winrt::name_of<$safeprojectname$::Scenario1_ShortName>()}
```

Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.cpp
```diff
@@ -6,16 +6,20 @@
...
-IVector<Scenario> implementation::MainPage::scenariosInner = winrt::single_threaded_observable_vector<Scenario>(
-	{
-		Scenario{ L"Scenario 1", winrt::hstring(winrt::name_of<$safeprojectname$::Scenario1_ShortName>())},
-		Scenario{ L"Scenario 2", winrt::hstring(winrt::name_of<$safeprojectname$::Scenario2_ShortName>())}
-	});
+namespace winrt::$safeprojectname$
+{
+    IVector<Scenario> implementation::MainPage::scenariosInner = winrt::single_threaded_observable_vector<Scenario>(
+        {
+            Scenario{ L"Scenario 1", winrt::hstring(winrt::name_of<$safeprojectname$::Scenario1_ShortName>())},
```
jonwis, line 19: Now that you're inside the `winrt` namespace (per line 15) you don't need the `winrt` prefixes inside it.  Name resolution will find them. So:

* `winrt::single_threaded_observable_vector` -> `single_threaded_observable_vector`
* `winrt::hstring` -> `hstring` (if still necessary)
* `winrt::name_of<>` -> `name_of<>`

Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/App.xaml.cpp
oldnewthing: @jonwis There are two issues here, and the OS samples repo is wrong on one count, possibly both. Fixing them is something I do only opportunistically, though.

First problem is the way we "slurk" the namespaces. Should be

```c++
using namespace winrt;
using namespace winrt::Windows::Foundation;
//              ^^^^^^^
```

If you leave off the `winrt::` prefix on `winrt::Windows::Foundation`, then you get a compile error if you try to use both C++/WinRT and WRL (which many people do).

```c++
#include <winrt/Microsoft.UI.Xaml.h>
#include <wrl/client.h>

using namespace winrt;
using namespace Microsoft::UI::Xaml; // error: "Microsoft" is ambiguous
// could be winrt::Microsoft or ::Microsoft
```

The second problem is whether we should be dumping everything into the global namespace, or consolidating in a `winrt` namespace:

```c++
// This is not controversial
namespace winrt
{
    using namespace Windows::Foundation;
}

// this part is
using namespace winrt;
```

For code that straddles multiple worlds, dumping everything into the global namespace is a bad idea because you want to be clearer about which `DispatcherQueue` you are talking about.

* `winrt::Windows::System::DispatcherQueue`
* `Windows:System::DispatcherQueue` (C++/CX dumps in the global namespace, alas)
* `ABI:Windows::System::DispatcherQueue`

The OS samples get away with `using namespace winrt;` because they live in only one world (C++/WinRT).

The WindowsAppSDK samples might or might not fall into the same category. If they need to juggle multiple projections and namespaces, then consolidating into a `winrt` non-global namespace may be better.

```c++
winrt::DispatcherQueue queue = blah blah; // Aha, this is the C++/WinRT version
ABI::DispatcherQueue* abi_queue = winrt::get_abi(queue); // convert to raw ABI
```
```

Comment (angelazhangmsft) [reply Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/App.xaml.cpp:9]: To be conservative and allow for calling other namespaces with WinAppSDK, does it sound ok to keep the pattern here? (e.g. global winrt namespace)?

File: Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp (hunk 2 above)
Comment (angelazhangmsft) [reply Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/MainWindow.xaml.cpp:86]: Will rename to `rc`. For context, these are the docs that the code is copied from: https://docs.microsoft.com/en-us/windows/win32/gdi/positioning-objects-on-a-multiple-display-setup

File: Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.cpp (hunk 3 above)
Comment (angelazhangmsft) [reply Templates/VSIX/Project Templates/cpp-winui-template/WinUI3TemplateCpp/SampleConfiguration.cpp:20]: When I tried that, I got the error: 
E0312	no suitable user-defined conversion from "std::basic_string_view<wchar_t, std::char_traits<wchar_t>>" to "winrt::hstring" exists	

Probably because `Project.idl` defines Scenario struct as follows:
```idl
struct Scenario
{
     String Title;
     String ClassName;
    };
```

PR #124, opened 2021-11-11
Samples VSIX Build And Release Pipeline

kythant:
VSIX Build and Release pipeline.
This PR defines the yml file that builds the Samples VSIX from a version of WindowsAppSDK and also creates the GitHub Release for it.
The github release task documented below is used to create the release and the parameters on the pipeline flows into the parameters of this task. 
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/github-release?view=azure-devops

**Parameters:**
_Release Unique Identifier_
This is used as a unique identifier for the GitHub Release.
Spaces not permitted

_GitHub Release Action_
- 'none'   
 This option only builds the VSIX, publishes it to Build Artifacts.
- 'create'
This option builds and publishes the VSIX and also creates a github release
- 'edit'
This option builds and publishes the VSIX and edits an existing github release
To edit an old release, provide "Release Unique Identifier" of the release you want to edit in the "Release Unique Identifier" field.
- 'delete'
This option does not build the VSIX and deletes an existing github release
To delete an old release, provide "Release Unique Identifier" of the release you want to delete in the "Release Unique Identifier" field.

_ReleaseTitle_
The title of the Release

_ReleaseNotes_
The notes of the Release

_IsDraft_
Indicates whether the release should be saved as a draft (unpublished). If unchecked, the release will be published

_IsPreRelease_
Indicate whether the release should be marked as a pre-release. 

_WinAppSDKVersion_
Version of the WindowsAppSDK used to build the Samples VSIX

### DIFF
--- a/WindowsAppSDK-BuildVSIX.yml
+++ b/WindowsAppSDK-BuildVSIX.yml
@@ -25,7 +25,7 @@ stages:
           Write-Host $WinAppSDKVersion
           [xml]$props = Get-Content -Encoding utf8 -Path $file
           $props.Project.PropertyGroup.WindowsAppSdkVersion.innerText = $WindowsAppSDKNugetPackageVersion
-          Write-Host $publicNuspec.OuterXml
+          Write-Host $publicNuspec
           Set-Content -Encoding utf8 -Value $publicNuspec.OuterXml $file
       
     - task: VSBuild@1
--- a/WindowsAppSDK-BuildVSIX.yml
+++ b/WindowsAppSDK-BuildVSIX.yml
@@ -17,7 +17,11 @@ parameters:
   - name: "ReleaseNotes"
     displayName: "Release Notes" 
     type: string
+    default: ''
   - name: "IsDraft"
+    type: boolean
+    default: False
+  - name: "IsPreRelease"
     type: boolean
     default: False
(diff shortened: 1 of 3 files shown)


WindowsAppSDK-BuildVSIX.yml
```diff
@@ -0,0 +1,110 @@
+# see https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases for info on yaml ADO jobs
+name: $(BuildDefinitionName)_$(date:yyMM).$(date:dd)$(rev:rrr)
+parameters:
+  # Version of the WindowsAppSDK used to build the Samples VSIX
+  - name: "WinAppSDKVersion"
+    displayName: "Windows App SDK version" 
+    type: string
+    default: '1.0.0-preview3'
```
angelazhangmsft, line 8: do we need to update this whenever WinAppSDKVersion in Directory.build.props is updated?

kythant: I think the default probably shouldn't be set here. 

But the default value in Directory.build.props probably should be updated to the latest release?

angelazhangmsft (approved):

PR #144, opened 2021-12-07
Update samples to enable builds with several versions of Visual Studio

erik0686:
## Description

I'm updating the samples to use the common props file to enable builds in different versions of Visual studio
@angelazhangmsft if you know the correct people to ask for review, I'd be grateful if you add them or let me know who they are :) 

## Checklist

### DIFF
--- a/Samples/AppLifecycle/Activation/cpp-console-unpackaged/CppWinRtConsoleActivation/CppWinRtConsoleActivation.vcxproj
+++ b/Samples/AppLifecycle/Activation/cpp-console-unpackaged/CppWinRtConsoleActivation/CppWinRtConsoleActivation.vcxproj
@@ -45,9 +45,6 @@
   </ItemGroup>
   <PropertyGroup Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
-    <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
-    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)'=='Debug'" Label="Configuration">
--- a/Samples/AppLifecycle/Activation/cpp-winui-packaged/CppWinUiDesktopActivation/CppWinUiDesktopActivation/CppWinUiDesktopActivation.vcxproj
+++ b/Samples/AppLifecycle/Activation/cpp-winui-packaged/CppWinUiDesktopActivation/CppWinUiDesktopActivation/CppWinUiDesktopActivation.vcxproj
@@ -55,7 +55,6 @@
   </ItemGroup>
   <PropertyGroup Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
     <DesktopCompatible>true</DesktopCompatible>
   </PropertyGroup>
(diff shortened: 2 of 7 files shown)


angelazhangmsft:
Thanks @erik0686! I've added the sample authors to the review.

AdamBraden:
Are these changes something that can be added as part of the template, so future samples don't run into it?

angelazhangmsft:
> Are these changes something that can be added as part of the template, so future samples don't run into it?

It looks like we currently manually import the [Build.Common.Cpp.props](https://github.com/microsoft/WindowsAppSDK-Samples/blob/main/Samples/Build.Common.Cpp.props) file using a relative path.

I wonder if we can use something like [this](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#customize-c-builds) to create a Directory.build.props that will automatically set these properties for C++ projects? That way, we could add an import statement to the C++ templates that doesn't rely on a relative path.

Samples/AppLifecycle/Activation/cpp-console-unpackaged/CppWinRtConsoleActivation/CppWinRtConsoleActivation.vcxproj
```diff
@@ -44,9 +45,6 @@
   </ItemGroup>
   <PropertyGroup Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
```
angelazhangmsft, line 47: We should probably leave in the `PlatformToolset` property since the samples are available as individual zip downloads through the samples browser, for example: https://docs.microsoft.com/samples/microsoft/windowsappsdk-samples/applifecycle/

If folks download any of the samples as a zip file, the Build.Common.Cpp.props file won't be available

erik0686: I think in that case VS will alert you that you don't have the correct version installed and will ask you to do it. 
I feel strongly about not having it explicitly defined here, since having it currently breaks the builds done in the internal repo, and will continue to do so in future efforts of having CI builds or other pipeline runs. Do you have any advice on how to tackle this?

angelazhangmsft: I see! If VS alerts you then I would say we don't need to set the property then. Would be good to verify with one of the C++ samples in https://docs.microsoft.com/en-us/samples/browse/?expanded=windows&products=windows-app-sdk

erik0686: Good idea, I'll verify

erik0686: After trying, this is what VS shows right after opening the solution 
```C:\Users\er\Downloads\ResourceManagement\cpp-console-unpackaged\console_unpackaged_app.vcxproj : warning  : The build tools for Visual Studio 2010 (v100) cannot be found. To build using the Visual Studio 2019 (v142) build tools, either click the Project menu or right-click the solution, and then select "Retarget Solution". ```

erik0686:
@AdamBraden 
RE: >Are these changes something that can be added as part of the template, so future samples don't run into it?

Off the top of my head, I don't know. Hopefully the answer is yes. I could take a look into it. The intent of this PR though, is to leave samples in a state where we could start the effort of a CI build and internal pipeline builds, so I would argue to have your proposal done in a different PR :) thoughts?

Samples/AppLifecycle/Activation/cpp-console-unpackaged/CppWinRtConsoleActivation/CppWinRtConsoleActivation.vcxproj
angelazhangmsft: Is there any way to fallback on the VS 2022 version of the build tools? It seems confusing to have a message regarding Visusal Studio 2010?

erik0686: I agree. However, I think this is something that (at least my) visual studio sets. I don't have anything related to VS2010 installed and I still got that. For example:

image

kythant (approved):

PR #168, opened 2022-02-07
User/andreww msft010

andreww-msft:
Fixed quarantined samples (maybe)


Samples/nuget.config
```diff
@@ -10,6 +10,8 @@
     <packageSources>
     <clear />
         <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+        <add key="Internal" value="https://pkgs.dev.azure.com/microsoft/ProjectReunion/_packaging/Project.Reunion.nuget.internal/nuget/v3/index.json" />
+        <add key="Local" value="C:\_Data\Cobalt\Reunion" />
```
kythant, line 14: Seems like this line is causing issues in the pipeline

erik0686: Also, I think we should not be using the internal feed :)

erik0686:
Can we change the title of the PR? Other than that this looks good to me :)

erik0686 (approved):

erik0686:
Should there be a follow up PR to move this sample from the quarantine folder?

kythant:
Some of these Samples are still in the quarantined folder. We need to move them to the Samples folder so that they can be properly tested by the CI/PR pipeline.

PR #170, opened 2022-02-09
Fix bad sample paths in readme.md

Scottj1s:
samples had bad paths


AdamBraden:
Thought these were under a DynamicDependencies folder and there were multiple samples, I can't find when that change was made. But this looks correct for the current structure.

AdamBraden (approved):

PR #171, opened 2022-02-14
User/andreww msft011

andreww-msft:
## Description

Changed Program.Main from async to synchronous. This is because async Main prevents Narrator reading XAML elements in a WinUI project. Because Main is no longer async, I also changed the redirection decision logic to make the async call on another thread, and to wait for this to completed using a non-blocking wait method.


Samples/AppLifecycle/Instancing/cs-winui-packaged/CsWinUiDesktopInstancing/CsWinUiDesktopInstancing/Program.cs
```diff
@@ -65,60 +65,6 @@ public static void ReportInfo(string message)
             }
         }
 
-        private static async Task<bool> DecideRedirection()
```
aeloros, line 68: Please reorder the file so it can be correctly diffed.  It makes it very hard to differentiate between new and old code.  When I checked out your branch locally and used a more advanced diff, it shows a bunch of new unrelated code was added.

aeloros (changes requested):

PR #175, opened 2022-03-06
Teststing unpackaged WinUi App

loneursid:
The goal is to create an unpackage WinUI WindowsAppSDK sample to demonstrate new features coming in v1.1.

This sample is based on the WinUI template from the templates folder.

The call to MddBootstrapInitialize() fails. I've verified that I'm passing the correct version number (we get a different error when passing the wrong version) and I've added the lib to the linker inputs and the dll to the delayloaded dlls.
 
I call ActivationRegistrationManager::RegisterForStartupActivation() for a proof of concept. I just wanted to see if I could call into the WindowsAppSDK, the actual sample will call APIs that aren't available in the current nuget. The call fails with 8000000E A method was called at an unexpected time. This seems to point to the method not being called on the right thread. 

For the sample to launch, the four MSIX in the tools directory of the WindowsAppSDK package must be manually installed.


loneursid:
experiment completed

PR #192, opened 2022-04-20
Create a script to update version in each project

kythant:
Create a script to update version in each project. This will allow for builds to easily update the version for testing.

### DIFF
--- a/UpdateVersions.ps1
+++ b/UpdateVersions.ps1
@@ -4,12 +4,18 @@ Param(
     [string]$RootPath = ""
 )
 
-Write-Host "Updating"
 Get-ChildItem -Recurse packages.config -Path $PSScriptRoot | foreach-object {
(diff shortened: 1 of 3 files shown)


UpdateVersions.ps1
```diff
@@ -0,0 +1,40 @@
+Param(
+    [string]$WinAppSDKVersion = "",
```
erik0686, line 2: Where are we getting this variable from?

kythant: From the caller :)

erik0686: Sorry, I missed the %* in the cmd file

kythant: Specifically, if the caller was our nightly build, the nightly build will have the version variable that it used for packing the nuget stored somewhere

erik0686 (approved):

PR #204, opened 2022-05-07
Restart API samples, cpp and cs

hamza-usmani:
## Description

Samples for the restart feature in 1.1 Preview 1 and onwards. An extremely simple/straightforward example of how to call the restart method and handle error responses in both cpp (console) and cs (winui) solutions.

## Target Release

1.1 Preview 1.


Samples/AppLifecycle/Restart/cpp-console-unpackaged/cpp-console-unpackaged.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    AppRestartFailureReason restartError{ AppInstance::Restart(winrt::to_hstring(restartArgs)) };
+
+    switch (restartError)
+    {
+        case AppRestartFailureReason::RestartPending:
+            std::cout << "Another restart is currently pending.";
+            break;
+        case AppRestartFailureReason::InvalidUser:
+            std::cout << "Current user is not signed in or not a valid user.";
```
andreww-msft, line 56: This seems like a curious case - when would this happen? ...and is there any other reason you'd get InvalidUser?

hamza-usmani: Because we leverage requestrestartasync, which _does_ return the invalid user failure reason, we have to catch this case as well.

andreww-msft (approved):

PR #207, opened 2022-05-11
Add localpackages + Cleanup git logwarnings

DrusTheAxe:
Add support for localpackages/ for local testing using private/physically-local nugets
Cleanup some build warnings (some of Aggregator build's 101 warnings)


loneursid (approved):
Changes to Notification Samples look good.

PR #339, opened 2024-05-06
User/niklasb/drawingisland text2

niklasb-ms:
## Description

Add text rendering to DrawingIsland sample.

### DIFF
--- a/Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
+++ b/Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
@@ -6,7 +6,7 @@
 #include "DrawingIsland.g.h"
 #include "IslandFragmentRoot.h"
 #include "TextRenderer.h"
-#include "TextElement.h"
+#include "TextItem.h"
 
 namespace winrt::DrawingIslandComponents::implementation
 {
@@ -92,7 +92,7 @@ namespace winrt::DrawingIslandComponents::implementation
         winrt::com_ptr<IRawElementProviderFragment> GetFragmentInFocus() const override;
 
     private:
-        VisualElement* HitTestVisual(
+        VisualItem* HitTestItem(
             float2 const& point) const;
 
         void Accessibility_Initialize();
@@ -248,8 +248,8 @@ namespace winrt::DrawingIslandComponents::implementation
         struct 
         {
             winrt::VisualCollection Visuals{ nullptr };
-            std::vector<std::unique_ptr<VisualElement>> VisualElements;
-            VisualElement* SelectedVisual{ nullptr };
+            std::vector<std::unique_ptr<VisualItem>> VisualItems;
+            VisualItem* SelectedItem{ nullptr };
             winrt::SpriteVisual CurrentColorVisual{ nullptr };
             float2 Offset{};
         } m_items;
--- a/Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
+++ b/Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
@@ -5,6 +5,8 @@
 
 #include "DrawingIsland.g.h"
 #include "IslandFragmentRoot.h"
+#include "TextRenderer.h"
+#include "TextElement.h"
 
 namespace winrt::DrawingIslandComponents::implementation
 {
@@ -90,7 +92,7 @@ namespace winrt::DrawingIslandComponents::implementation
         winrt::com_ptr<IRawElementProviderFragment> GetFragmentInFocus() const override;
 
     private:
-        winrt::Visual HitTestVisual(
+        VisualElement* HitTestVisual(
             float2 const& point) const;
 
         void Accessibility_Initialize();
@@ -194,6 +196,8 @@ namespace winrt::DrawingIslandComponents::implementation
         {
             winrt::Compositor Compositor{ nullptr };
 
+            std::shared_ptr<TextRenderer> TextRenderer;
+
             // Current color used for new items
             unsigned int CurrentColorIndex = 0;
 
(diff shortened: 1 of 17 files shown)


Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
```diff
@@ -194,6 +196,8 @@ namespace winrt::DrawingIslandComponents::implementation
         {
             winrt::Compositor Compositor{ nullptr };
 
+            std::shared_ptr<TextRenderer> TextRenderer;
```
jeffstall, line 199: You're initializing text as part of UIA? What if UIA is never instantiated? This seems odd.

Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
```diff
@@ -244,7 +248,8 @@ namespace winrt::DrawingIslandComponents::implementation
         struct 
         {
             winrt::VisualCollection Visuals{ nullptr };
```
jeffstall, line 250: Since you're adding more members, can you start to add comments about what each of them to? For example, what's the different between Visuals and VisualItems? Do we want to just call "VisualItem" as "Item" to avoid any confusion?

Samples/Islands/DrawingIsland/DrawingIslandComponents/TextRenderer.cpp
```diff
@@ -0,0 +1,183 @@
...
+            defaultFontSize,
+            L"en-US",
+            m_textFormat.put()
+        ));
+        winrt::check_hresult(m_textFormat->SetWordWrapping(DWRITE_WORD_WRAPPING_NO_WRAP));
+
+        // Create the D2D factory.
+        winrt::check_hresult(D2D1CreateFactory(D2D1_FACTORY_TYPE_MULTI_THREADED, m_d2dFactory.put()));
+
+    }
```
jeffstall, line 35: Remove extra blank line above.

Samples/Islands/DrawingIsland/DrawingIslandComponents/TextRenderer.cpp
```diff
@@ -0,0 +1,183 @@
...
+        m_compositionGraphicsDevice = compositionGraphicsDevice.as<CompositionGraphicsDevice>();
+        m_d2dDevice = std::move(d2dDevice);
+    }
+
+    void TextRenderer::Render(
+        _In_z_ WCHAR const* text,
+        Windows::UI::Color backgroundColor,
+        Windows::UI::Color textColor,
+        SpriteVisual const& visual
+    )
```
jeffstall, line 106: Stylistic: For consistency throughout the sample, can you put the closing `)` at the end of the last line instead of on its own line?

Samples/Islands/DrawingIsland/DrawingIslandComponents/TextRenderer.cpp
```diff
@@ -0,0 +1,183 @@
...
+    {
+        constexpr float marginLeft = 5;
+        constexpr float marginRight = 5;
+        constexpr float marginTop = 5;
+        constexpr float marginBottom = 5;
+
+        winrt::com_ptr<IDWriteTextLayout> textLayout;
+        winrt::check_hresult(m_dwriteFactory->CreateTextLayout(
```
jeffstall, line 114: Can you add some comments here about why this is doing this? For example, is this how you measure the text so we know how large to make the backing surface?

Samples/Islands/DrawingIsland/DrawingIslandComponents/VisualItem.h
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#pragma once
+
+namespace winrt::DrawingIslandComponents::implementation
+{
+    class VisualItem
```
jeffstall, line 8: Please add a class comment that explains __what__ a VisualItem's primary purpose is for, and why it's not just a Visual. For example, this is how we notify owners when they need to handle a DPI scale change to re-rastersize their text.

jeffstall (approved):
Please address comments, then can check in.

Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
niklasb-ms: I don't know about UIA. I put the TextRenderer alongside the Compositor because it depends on the compositor and is used to create visuals for text items. It seemed like a reasonable fit, but I don't really understand the purpose of grouping the m_output members into a struct, so I could easily be wrong about that. I'm happy to move it if you can suggest another location.

Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
niklasb-ms: OK.

Samples/Islands/DrawingIsland/DrawingIslandComponents/TextRenderer.cpp
niklasb-ms: Please let me know if I missed any.

Samples/Islands/DrawingIsland/DrawingIslandComponents/DrawingIsland.h
jeffstall: Never-mind. Looking at the code in GitHub was confusing. I got it successfully to load in CodeFlow, which is much better.

jeffstall (approved):
:shipit:

jeffstall:
Thanks!

---
In reply to: [2042139403](https://github.com/microsoft/WindowsAppSDK-Samples/pull/339#issuecomment-2042139403) [](http://example.com/codeflow?ancestors=2042139403)

jeffstall:
Looks good!

PR #341, opened 2024-05-07
Draw the background color visual on top of the item visuals.

niklasb-ms:
## Description

Change order of the visuals in the DrawingIslands sample such that the visual showing the background color (the bar across the bottom of the window) is on top.


jeffstall (approved):
Looks good!

PR #346, opened 2024-05-09
Update DrawingCsTestApp's SceneNode scenario to load the helmet model

achen144:
Updates DrawingCsTestApp's SceneNode scenario to load the helmet model into Scene API objects, with a rotation animation applied. It reads the model's textures and mesh data from asset files, instead of hard-coded binary blobs.


jeffstall:
Good

PR #348, opened 2024-05-09
Add Mica (Win11) and Acrylic (Win10) backdrops

adpa-ms:
Added SystemBackdropControllers to draw Mica (Win11+) and Acrylic (Win10) behind the DrawingIsland.


jeffstall (approved):
Thanks!

PR #351, opened 2024-05-09
Handle device-lost for text in DrawingIslands sample.

niklasb-ms:
## Description

This enables the text rendering in the `DrawingIslands` sample to properly recover when the D3D device is lost.

The `DeviceLostHelper` class listens for device-lost events on the specified DXGI device. This code is adapted from https://learn.microsoft.com/en-us/windows/uwp/composition/composition-native-interop.

The `DrawingIsland` class now includes an instance of `DeviceLostHelper` and uses it to listen for events on the device associated with the `TextRenderer` object. On device lost, the `DrawingIsland::OnDirect3DDeviceLost` method is called and does the following:

1. Tells the `TextRenderer` to recreate its device
2. Re-renders all the text items
3. Starts watching the new device

The `TextRenderer` class has new code to recreate its device. It also catches and ignores the exception in the unlikely event that a DXGI_ERROR_DEVICE_REMOVED error occurs _during_ `TextRenderer::Render`.

I tested this by triggering multiple device losses using `dxcap -forceTDR`.


jeffstall (approved):
:shipit:

jeffstall:
Thanks! Right now, this introduces multi-threading in, which gets us unblocked. I'll refactor this around later to use a DispatcherQueue to post back to the UI thread.

PR #352, opened 2024-05-09
Marshal OnDirect3DDeviceLost to UI thread.

niklasb-ms:
## Description

This change fixes a thread safety issue with DrawingIsland::OnDirect3DDeviceLost. This method recreates the D2D device and re-renders text items after device lost. The problem is that this method is called from a threadpool thread, not the UI thread. The fix is to use DispatcherQueue to marshal the call to the UI thread.

I tested this by forcing device loss using `dxcap -forceTDR` and verifying that text still renders correctly. I also verified in the debugger that `TextRenderer::RecreateDirect2DDevice` is called from the UI thread.


jeffstall (approved):
:shipit:

jeffstall:
Thanks!

PR #384, opened 2024-09-04
Remove PublishReadyToRun from pubxml to allow Override from MSBuild cmd

kythant:
We are setting PublishReadyToRun property to false to fix a nuget restore issue on the package that depends on it however the PublishReadyToRun property in the publxml is overriding everything. Hence, these properties should be moved to the csprojs


kythant:
Test Successful here
https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=106299976

alexlamtest (approved):

JesseCol:
Non-blocking: you can delete the template text in the description.  I want to change this since it's obsolete now but I haven't gotten around to it :)

kythant:
> Non-blocking: you can delete the template text in the description. I want to change this since it's obsolete now but I haven't gotten around to it :)

Done!

PR #417, opened 2025-03-17
Add arm64 config step to WCR samples

niels9001:
Small, but crucial, step in building the WCR samples.


pratikone (approved):
wish we could make it default in VS somehow,  but good call.

PR #431, opened 2025-03-31
Add a sample of a .NET MAUI app using WCR

codendone:
## Description

This adds a sample which shows how to use Windows Copilot Runtime (and Windows App SDK in general) in a .NET MAUI application.

## Target Release

Whenever WindowsCopilotRuntime gets to "stable". This currently targets 1.7-experimental3.

## Checklist


Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/MauiWindowsCopilotRuntimeSample.sln
```diff
@@ -0,0 +1,27 @@
...
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MauiWindowsCopilotRuntimeSample", "MauiWCRSample\MauiWindowsCopilotRuntimeSample.csproj", "{5B91DC6A-CF3C-4398-BFF3-A45B88BFA698}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{5B91DC6A-CF3C-4398-BFF3-A45B88BFA698}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
jeffstall, line 14: Checking: What does "Any CPU" mean with WinAppSDK? For our Build/Ignite talks, we guided people to replace with explicit use for x86, x64, and arm64. Does MAUI abstract this out so they can safely use "Any CPU" instead?

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/Platforms/Windows/MainPage.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        partial void ScaleImage(double scale)
+        {
+            DoScaleImage(scale);
+        }
+
+        private async void DoScaleImage(double scale)
+        {
+            // Load the original image
+            var resourceManager = new Microsoft.Windows.ApplicationModel.Resources.ResourceManager();
```
jeffstall, line 55: Since your project has Android and other platforms, does this need to be under some WINDOWS_ONLY check or is this already abstracted out? i.e., with the other platform builds fail at compile/runtime? Should there be some IsWindows() runtime check or something?

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/Platforms/iOS/Program.cs
```diff
@@ -0,0 +1,15 @@
+﻿using ObjCRuntime;
+using UIKit;
+
+namespace MauiWindowsCopilotRuntimeSample;
+
+public class Program
+{
+	// This is the main entry point of the application.
```
jeffstall, line 8: Tabs as spaces?

jeffstall (approved):
Added some comments for clarification/double checking, but not blocking check-in.

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/MauiWindowsCopilotRuntimeSample.csproj
```diff
@@ -0,0 +1,68 @@
...
+
+		<!-- Versions -->
+		<ApplicationDisplayVersion>1.0</ApplicationDisplayVersion>
+		<ApplicationVersion>1</ApplicationVersion>
+
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">11.0</SupportedOSPlatformVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'">13.1</SupportedOSPlatformVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">21.0</SupportedOSPlatformVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</SupportedOSPlatformVersion>
+		<TargetPlatformMinVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</TargetPlatformMinVersion>
```
KevinLaMS, line 37: 17763 cannot be correct for line 36 and 37

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/MauiWindowsCopilotRuntimeSample.sln
codendone: This surprised me as well. Like you said, for WinAppSDK I think we say that an explicit architecture needs to be used, or at least there need to be platform-specific versions for apps uploaded to the store. A default MAUI project is configured only for "Any CPU", so I'm assuming they have some abstraction. I've chosen to leave this untouched from the default MAUI project.

Samples/WindowsCopilotRuntime/cs-maui/README.md
```diff
@@ -0,0 +1,60 @@
...
+extendedZipContent:
+- path: LICENSE
+  target: LICENSE
+---
+
+## Prerequisites
+- For system requirements, see [System requirements for Windows app development](https://docs.microsoft.com/windows/apps/windows-app-sdk/system-requirements).
+- To ensure your development environment is set up correctly, see [Install tools for developing apps for Windows 10 and Windows 11](https://docs.microsoft.com/windows/apps/windows-app-sdk/set-up-your-development-environment).
+- Running this sample does require a [Windows Copilot + PC](https://learn.microsoft.com/windows/ai/npu-devices/)
+- This sample requires [Windows App SDK 1.7 Experimental3](https://learn.microsoft.com/windows/apps/windows-app-sdk/downloads#windows-app-sdk-17-experimental), which should be installed by Visual Studio when deploying or launching the app.
```
KevinLaMS, line 20: How do we remember to bump this when we release?

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/Platforms/Windows/MainPage.cs
codendone: It is true that `#if WINDOWS` blocks could be used to ensure code in those blocks are only used in the build for Windows target. But I found that approach to be a poor choice for anything more than a line or two of trivial code, since that code is gray-ed out (VS sees the #define isn't always set) and there is no IntelliSense/auto-suggest inside those blocks.

What this sample uses instead is MAUI's other platform-specific code option: defining partial classes with partial methods and implementing platform-specific code in the platform-specific versions of those partial classes. This `Platforms/Windows/MainPage.cs` is the Windows-specific implementation of the partial class, so no `#if WINDOWS` lines are needed.

Samples/WindowsCopilotRuntime/cs-maui/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+This sample demonstrates use of some Windows Copilot Runtime APIs, including `LanguageModel` for text
+generation and `ImageScaler` for image super resolution to scale and sharpen images. Click one of the
+"Scale" buttons to scale the image (or reshow the original, unscaled image), or enter a text prompt and
+click the "Generate" button to generate a text response.
+
+The changes from the ".NET MAUI App" template are split across four files:
+- `MauiWindowsCopilotRuntimeSample.csproj`: Adds the required Windows App SDK package reference for the
+  Windows Copilot Runtime APIs. This reference needs to be conditioned only when building for Windows
+  (see below for details). This file also sets the necessary TargetFramework for Windows.
```
KevinLaMS, line 39: to be more prescriptive... should it be see Additional notes below?

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/MauiWindowsCopilotRuntimeSample.csproj
codendone: It is correct. The sample "works" on 17763+. The WCR functionality isn't available there, of course, and those portions of the UI will be hidden if WCR is not available.

LanceMcCarthy:
The nuget scenario here doesn't seem to work for anyone outside of how you have your `localpackages` set up. I am recommending a small update to the shared nuget.config file to allow supporting anyone who clones the code (and it still supports your AzDO/GHA restores)

Update the entire contents of `WindowsAppSDK-Samples/Samples/nuget.config`

to be this (I added comments to explain further

```xml
<?xml version="1.0" encoding="utf-8"?>
<!-- A dedicated nuget.config to provide grerater control over builds in CI and locally -->
<configuration>
    <packageRestore>
        <!-- Allow NuGet to download missing packages -->
        <add key="enabled" value="True" />
        <!-- Automatically check for missing packages during build in Visual Studio -->
        <add key="automatic" value="True" />
    </packageRestore>
    <packageSources>
        <!-- Clear any local or duplicate feeds. This gives us precise control over the sources used. -->
        <clear />
        
        <!-- Start with nuget.org source -->
        <add key="NugetOrg" value="https://nuget.org/v3/index.json" />

        <!-- Add the demo-specific feed, which may contain more recent packages not yet available on nuget.org -->
        <add key="WinAppSDK-SampleDeps" value="https://pkgs.dev.azure.com/shine-oss/WinAppSDK-Samples/_packaging/WinAppSDK-SampleDeps/nuget/v3/index.json" />

        <!-- a local directory to allow testing nupkg files without pushing to a remote feed -->
        <!-- This is also used by WinAppSDK's pipelines to validate a nupkg from a build -->
        <add key="localpackages" value="localpackages" />
    </packageSources>
    <disabledPackageSources />
</configuration>

```

Samples/WindowsCopilotRuntime/cs-maui/README.md
codendone: Currently the WCR samples are only in the `release/experimental` branch. In order to move them to the `main` branch (once WCR goes stable), we'll need to update them to point to the minimum stable build. After that, we theoretically bump all samples in `main` to the latest release after each stable release (not the monthly releases).

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/MauiWindowsCopilotRuntimeSample.sln
LanceMcCarthy: Since I've already chimed in on this PR, I thought I might add some helpful comments ;)

This is the correct solution configuration for a .NET MAUI single-project. When you have the netX.0-windows TFM selected, MAUI will automatically choose the current platform as the RID override.

image

Side Note:
It is a different story when it comes to publishing. You must explicitly choose the RID manually in your `dotnet publish` command using the `RuntimeIdentifierOverride` param. For example:

`dotnet publish ${{env.CSPROJ_PATH}} -f "net9.0-windows10.0.22621.0" -p:RuntimeIdentifierOverride="win10-x64" -p:PublishTrimmed=False -p:GenerateAppxPackageOnBuild=true -p:AppxPackageSigningEnabled=false -p:EnableMsixTooling=true --self-contained=true`

Samples/WindowsCopilotRuntime/cs-maui/MauiWCRSample/Platforms/Windows/MainPage.cs
LanceMcCarthy: It doesn't _have_ to be a partial class, any code files you put inside the `Platform/PlatformName` folder, it wil automatically be compiled only for that target platform. Functionally identical to using `#IF platformname` anywhere else in the project..

Using a partial class for MainPage.xaml.cs just makes it more convenient to access those methods... however, I urge a little bit of caution here. It isn't obvious that the logic is within the Platform/Windows folder, so I would add a comment to the original MainPage.xaml/MainPage.xaml.cs to the reader to let them know where to look for the rest of the partial class.

codendone:
> The nuget scenario here doesn't seem to work for anyone outside of how you have your `localpackages` set up...

It looks like the main branch has a nuget.org reference, though in a different order. I'll follow-up on this separately.

PR #442, opened 2025-04-11
Fix WCRForWPF Sparse Pacakge App crashing on ImageBuffer.CreateCopyFromBitmap(softwareBitmap);

kythant:
## Description
The AppxManifest needs to define the dependency on the WindowsAppRuntime.

I've also implemented an -clean option on the Build script to ensure that the MSIX will get replaced too.


codendone (approved):
:shipit:

PR #447, opened 2025-04-18
[WCR Sample] Single dotnet run with unpackage selfcontained as example.

yeelam-gordon:
## Description

Given WCR already support Self-contained in WinAppSDK 1.8 experimental 1, updated the dotnet run command example in readme accordingly.

## Target Release

Please specify which release this PR should align with WCR WinAppSDK 1.8 experimental 1, which is what the project is referencing now.

### DIFF
--- a/Samples/WindowsCopilotRuntime/README.md
+++ b/Samples/WindowsCopilotRuntime/README.md
@@ -39,17 +39,12 @@ join the [Windows Insider Program](https://insider.windows.com).
 # Contributing to this project
 - Refer to the [contributing guide](./Contributing.md)
 
-## Special Considerations for Debugging as an Unpackaged App in Visual Studio
+## Special Considerations for Unpackaged & Self-Contained Mode with WCR
 
-- This project is not designed to be fully functional in an unpackaged app. However, Windows Copilot Runtime does support unpackaged app.
-- To enable proper startup as an unpackaged app, you need to bootstrap the Windows App SDK either Programmatically or By adding the following configuration to the `.csproj` file during the build process:
- ```xml
-  <WindowsAppSdkBootstrapInitialize>true</WindowsAppSdkBootstrapInitialize>
-  <WindowsPackageType>None</WindowsPackageType>
- ```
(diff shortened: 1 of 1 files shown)


Samples/WindowsCopilotRuntime/README.md
```diff
@@ -39,17 +39,12 @@ join the [Windows Insider Program](https://insider.windows.com).
...
-- This project is not designed to be fully functional in an unpackaged app. However, Windows Copilot Runtime does support unpackaged app.
-- To enable proper startup as an unpackaged app, you need to bootstrap the Windows App SDK either Programmatically or By adding the following configuration to the `.csproj` file during the build process:
- ```xml
-  <WindowsAppSdkBootstrapInitialize>true</WindowsAppSdkBootstrapInitialize>
-  <WindowsPackageType>None</WindowsPackageType>
- ```
-- Alternatively, if you are using the Developer Command Prompt for Visual Studio, you can run the app as an ARM64 version using the following command, with the bootstrap property provided:
+## Special Considerations for Unpackaged & Self-Contained Mode with WCR
+
+- This project is not intended to be fully functional when running as an unpackaged app. However, Windows Copilot Runtime (WCR) does support unpackaged apps.
```
pratikone, line 44: can this be changed to, by default, this project runs as packaged, but you can also run it unpackaged by setting <WindowsPackageType>None </WindowsPackageType> in .csproj

pratikone (approved):

Samples/WindowsCopilotRuntime/README.md
yeelam-gordon: Talked with Pratik offline.
We removed the paragraph to talk about adding the property as part of csproj for unpackage app, and here is the reason:
a. It is over what this Readme should have, and there are multiple option to build/debug as unpackage app in VS which we will not repeat here.
b. Our goal is to give a single command, which in case people don't have VS and not yet install WinAppSDK, they can still able to use dotnet run to launch this sample directly.

PR #468, opened 2025-05-20
Merging WCR /WinAI API sample from release/experimental branch

pratikone:
- All new experimental api changes
-  new name branding and namespace change
- README and contributing doc moved inside cs-winui because other samples will also exist within this folder.


pratikone:
not merging this. making another PR

PR #509, opened 2025-09-05
Use the new Microsoft.Windows.Storage.Pickers API

DinahK-2SO:
Use the new Microsoft.Windows.Storage.Pickers API from WindowsAppSDK.1.8

### DIFF
--- a/Samples/WindowsAIFoundry/cs-winui/ViewModels/InputImageViewModelBase.cs
+++ b/Samples/WindowsAIFoundry/cs-winui/ViewModels/InputImageViewModelBase.cs
@@ -48,12 +48,12 @@ internal abstract class InputImageViewModelBase<T> : CopilotModelBase<T>
 
         _pickInputImageCommand = new(async _ =>
         {
-            var picker = new FileOpenPicker(App.window.Id);
-            picker.ViewMode = PickerViewMode.Thumbnail;
-            picker.SuggestedStartLocation = PickerLocationId.PicturesLibrary;
-            picker.FileTypeFilter.Add(".jpg");
-            picker.FileTypeFilter.Add(".jpeg");
-            picker.FileTypeFilter.Add(".png");
+            var picker = new FileOpenPicker(App.window.Id)
+            {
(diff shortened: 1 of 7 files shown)


Samples/WindowsAIFoundry/cs-winui/ViewModels/InputImageViewModelBase.cs
```diff
@@ -48,18 +48,20 @@ await DispatcherQueue.EnqueueAsync(async () =>
...
-            InitializeWithWindow.Initialize(picker, hwnd);
-
-            picker.ViewMode = PickerViewMode.Thumbnail;
-            picker.SuggestedStartLocation = PickerLocationId.PicturesLibrary;
-            picker.FileTypeFilter.Add(".jpg");
-            picker.FileTypeFilter.Add(".jpeg");
-            picker.FileTypeFilter.Add(".png");
-
-            var file = await picker.PickSingleFileAsync();
+            var picker = new FileOpenPicker(App.window.Id)
```
yeelam-gordon, line 51: I think no compliable here, since i tried it.
Even you change to App.Window.AppWindow.Id will not be working in runtime, since i tried it.

yeelam-gordon: I can put my version in your PR here.
Basically, beside using our picker, I also get rid of StorageFile in below usage.

yeelam-gordon: Sorry, @DinahK-2SO, I found that I cannot apply the file pickers here.
Since 1.8 stable build doesn't have all API need for previous 1.8 experiemental release for this AI project.
And our picker in 1.8 exp4 doesn't work here as well because of localization problem.

DinahK-2SO: My local build has the same issue. I am converting this PR to draft. Let's wait for the samples to migrate to official release.

PR #524, opened 2025-09-22
Download SqueezeNet model and image files rather than storing them in the repository

nieubank:


Samples/WindowsML/Resources/SqueezeNetModel.targets
```diff
@@ -0,0 +1,72 @@
...
+    <Message Text="Checking for SqueezeNet model in: $(SqueezeNetModelTempDir)" Importance="high" />
+    
+    <!-- Download model file if it doesn't exist -->
+    <DownloadFile 
+      SourceUrl="$(SqueezeNetModelUrl)" 
+      DestinationFolder="$(SqueezeNetModelTempDir)" 
+      DestinationFileName="SqueezeNet.onnx"
+      Condition="!Exists('$(SqueezeNetModelFile)')"
+      ContinueOnError="true" />
+    <Warning Text="SqueezeNet model download may have been interrupted by antivirus scanning or network issues. The file may become available shortly as antivirus completes its scan."
```
adrastogi, line 36: >Warning [](http://example.com/codeflow?start=5&length=7)

Looking at the task documentation, it sounds like it has retry and delay functionality which could be considered to further harden against this. [https://learn.microsoft.com/en-us/visualstudio/msbuild/downloadfile-task?view=vs-2022](https://learn.microsoft.com/en-us/visualstudio/msbuild/downloadfile-task?view=vs-2022)

PR #529, opened 2025-09-23
Add and Call InitializeProviders in GenAI samples.

chrisdMSFT:
- Add and call to `InitializeProvidersAsync` in the HelloPhi sample.
- Add and caIl `InitializeProviders` in CppConsoleDesktop.GenAI sample.


Samples/WindowsML/cs/HelloPhi/Program.cs
```diff
@@ -72,6 +113,8 @@ static void PrintUsage()
 int minLength = 50;
 int maxLength = 500;
 
+
```
adrastogi: Nit: extra whitespace?

PR #530, opened 2025-09-23
Improvements to WindowsML sample for MSIX packaging

nieubank:
- Add SqueezeNet model files for MSIX sample
- Remove unpackaged profile from launch settings for WindowsML sample


Samples/WindowsML/cs-winui/WindowsMLSample.csproj
```diff
@@ -28,6 +28,27 @@
     </Content>
   </ItemGroup>
 
+  <!-- Include SqueezeNet model files in MSIX package -->
+  <Target Name="IncludeSqueezeNetFilesInPackage" AfterTargets="CopySqueezeNetModelToOutput" BeforeTargets="_ComputeAppxPackagePayload">
```
manodasanW, line 32: Does Content include not work similar to the above with jpg / png for assets?

nieubank: If we put these in the Assets folder, it may work, specifying them as Content did not enable them to do into the root directory where the sample code is expecting them though.

PR #534, opened 2025-09-24
Windows ML: Consolidate ResNet projections into main solution and fix warnings

adrastogi:
## Description
The ResNet-related projects previously had their own solutions; we had cleaned up the one for C++ but left behind the C# one. This change removes that solution and then adds both projects to the main solution for the Windows ML area. This change also fixes a couple warnings Visual Studio flagged in the code. 

## Target Release

1.8.1

## Checklist


Samples/WindowsML/cs-winui/MainWindow.xaml.cs
```diff
@@ -135,7 +135,8 @@ private async Task LoadModelAsync()
         private async Task LoadAndRunDemoAsync()
         {
             // Get application directory
-            var appDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var appDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)
```
manodasanW, line 138: Consider if this works for your scenarios or not: https://learn.microsoft.com/en-us/dotnet/api/system.appcontext.basedirectory?view=net-9.0

adrastogi: Thanks! That looks like a nicer approach- I've noted this for subsequent updates but will leave as-is for now.

PR #541, opened 2025-10-01
Add ModelCatalog samples to WinML

dabhattimsft:
## Description

Adding WinML model catalog sample code to existing WinML samples. Adding new --use_model_catalog command line option, which will use SqueezeNetModelCatalog.json to download squeezenet model files and use those instead of using local copies.

## Target Release

2.0

## Checklist


Samples/WindowsML/Resources/SqueezeNetModelCatalog.json
```diff
@@ -0,0 +1,67 @@
...
+      "description": "",
+      "modelSizeBytes": 1250000,
+      "license": "BSD",
+      "licenseUri": "https://github.com/microsoft/WindowsAppSDK-Samples/raw/refs/heads/release/experimental/Samples/WindowsML/cpp/CppConsoleDesktop/CppConsoleDesktop/SqueezeNet.LICENSE.txt",
+      "licenseText": "This model is provided under the License Terms available at \u003Chttps://github.com/microsoft/WindowsAppSDK-Samples/raw/refs/heads/release/experimental/Samples/WindowsML/cpp/CppConsoleDesktop/CppConsoleDesktop/SqueezeNet.LICENSE.txt\u003E",
+      "uri": "https://github.com/microsoft/WindowsAppSDK-Samples/raw/refs/heads/release/experimental/Samples/WindowsML/cpp/CppConsoleDesktop/CppConsoleDesktop",
+      "files": [
+        {
+          "name": "SqueezeNet.onnx",
+          "uri": "https://github.com/microsoft/WindowsAppSDK-Samples/raw/refs/heads/main/Samples/WindowsML/Resources/SqueezeNet.onnx",
```
nieubank: This should use a permalink?

Samples/WindowsML/Shared/cs/ArgumentParser.cs
```diff
@@ -179,6 +184,7 @@ public static void PrintHelp()
             Console.WriteLine("  --device_type <type>        Optional hardware device type to use when EP supports multiple (e.g. CPU, GPU, NPU)");
             Console.WriteLine("  --compile                   Compile the model");
             Console.WriteLine("  --download                  Download required packages");
+            Console.WriteLine("  --use_model_catalog         Use the model catalog for model discovery");
             Console.WriteLine("  --model <path>              Path to input ONNX model (default: SqueezeNet.onnx)");
```
nieubank, line 188: Should we make model catalog the default unless an explicit path is provided?

nieubank (approved):

Samples/WindowsML/Resources/SqueezeNetModelCatalog.json
```diff
@@ -0,0 +1,67 @@
+{
+  "base": "https://learn.microsoft.com/azure/ai-foundry/foundry-local/reference/reference-catalog-api",
+  "models": [
+    {
+      "alias": "squeezenet",
+      "name": "squeezenet",
+      "version": "1",
+      "modelType": "ONNX",
+      "publisher": "microsoft",
+      "executionProvider": "CPUExecutionProvider",
```
adrastogi, line 10: Can you remind me how the implementation uses the EP info? Technically, I believe this model can run on the different NPUs, so would this be more accurately described via an array of those?

Samples/WindowsML/cpp/CppConsoleDesktop/CppConsoleDesktop.vcxproj
```diff
@@ -1,25 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Enable WinML header include path -->
+  <Import Project="..\..\packages\Microsoft.WindowsAppSDK.2.0.250918002-experimental1\build\native\Microsoft.WindowsAppSDK.props" Condition="Exists('..\..\packages\Microsoft.WindowsAppSDK.2.0.250918002-experimental1\build\native\Microsoft.WindowsAppSDK.props')" />
```
adrastogi: I think we may need to keep the macro that was there previously. I believe the WASDK folks use it for central override, among other things, possibly. 

(When you were updating the package versions, did you do so through VS's package management UI? I know that flow can take care of efficiently updating the project files but probably does not preserve the macro. Not sure if there's a better way to manage this.)

dabhattimsft: $(NugetPackageDirectory)?

I did use VS

Samples/WindowsML/Resources/SqueezeNetModelCatalog.json
dabhattimsft: We get EPs from ExecutionProviderCatalog::GetDefault().FindAllProviders() and filter models based on that. I can put array but for sample purposes I just kept cpu

Samples/WindowsML/Shared/cs/ArgumentParser.cs
dabhattimsft: Since it will download the model, local model is faster. Hence kept it optional. May revisit that when we GA.

Samples/WindowsML/cpp/CppConsoleDesktop/CppConsoleDesktop.vcxproj
adrastogi: Tagging @manodasanW and @Scottj1s for the parenthetical in the above comment; can you help me clarify the scenarios for having the packages path prefixed with `$(NugetPackageDirectory)`? And if the recommendation is to prefer that mechanism vs. a relative path, is there a smooth way to manage package updates? @dabhattimsft did the natural thing and used Visual Studio's UX to update packages, but that stomps on the macro.

Scottj1s: yeah, unfortunately, packages.config (vcxproj) insists on solution-specific package directories, vs PackageReference (csproj) use of a global shared cache.  some teams work around that by editing the project to use a variable, and the pipeline also passes that into the build script.  (another technique would be to set repositoryPath in nuget.config, but that's not as flexible an approach).  we should maintain NugetPackageDirectory in samples.

Samples/WindowsML/Resources/SqueezeNetModelCatalog.json
adrastogi: Thanks for the quick offline chat. A nice future addition would be to expand this to support other EP types to illustrate how that would be done and then to also enable running the catalog scenario against specific EPs.

PR #559, opened 2025-10-15
Windows ML: Add runtime ID to C# console sample

adrastogi:
## Description

#556 introduced some changes to make the C# console desktop application self-contained. I saw build breaks in the pipeline though due to mismatches between the platform and the runtime (self-contained requires a strict alignment between the two):

`C:__t\dotnet\sdk\8.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(288,5): error NETSDK1032: The RuntimeIdentifier platform 'win-x64' and the PlatformTarget 'arm64' must be compatible. [C:__w\1\s\WindowsAppSDK-Samples\Samples\WindowsML\cs\CSharpConsoleDesktop\CSharpConsoleDesktop\CSharpConsoleDesktop.csproj]`

This change addresses the problem by specifying the runtime identifier based on the supplied platform. I tested this via `dotnet build` and supplying the corresponding versions, which seemed to work, and I'll further validate in the pipeline itself. 

## Target Release

1.8.3

## Checklist

### DIFF
--- a/Samples/WindowsML/cs/CSharpConsoleDesktop/CSharpConsoleDesktop/CSharpConsoleDesktop.csproj
+++ b/Samples/WindowsML/cs/CSharpConsoleDesktop/CSharpConsoleDesktop/CSharpConsoleDesktop.csproj
@@ -4,6 +4,8 @@
     <OutputType>Exe</OutputType>
     <WindowsPackageType>None</WindowsPackageType>
     <SelfContained>True</SelfContained>
+    <RuntimeIdentifier Condition="'$(Platform)' == 'x64'">win-x64</RuntimeIdentifier>
+    <RuntimeIdentifier Condition="'$(Platform)' == 'ARM64'">win-arm64</RuntimeIdentifier>
   </PropertyGroup>
 
   <ItemGroup>


Samples/WindowsML/cs/CSharpConsoleDesktop/CSharpConsoleDesktop/CSharpConsoleDesktop.csproj
```diff
@@ -4,6 +4,8 @@
     <OutputType>Exe</OutputType>
     <WindowsPackageType>None</WindowsPackageType>
     <SelfContained>True</SelfContained>
+    <RuntimeIdentifier Condition="'$(Platform)' == 'x64'">win-x64</RuntimeIdentifier>
```
mschofie, line 7: I haven't seen RuntimeIdentifier be conditional before - it's normally specified as a set, like:

```xml
<RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>
```

Do you have to constrain it by platform?

adrastogi: Yeah, good observation- I'm used to seeing them declared that way as well. And that is actually the approach that I tried originally, but it did not resolve the problem I'm seeing in the build pipeline (I got back essentially the same error where inapplicable RuntimeIdentifier / Platform values are being paired up, e.g. arm64 and win-x64). I couldn't figure out how to prevent that from happening- so, definitely open to feedback on how to write this more idiomatically while ensuring the right combinations are used. 

I'm wondering if the SelfContained aspect of this sample has anything to do with it- the other C# samples use the expected "phrasing" of the Platform / RuntimeIdentifier which is in the Directory.Build.props file, and those don't seem to hit this problem (at least, I haven't seen evidence yet). But none of those programs are SelfContained.

mschofie: It's possible that because the build always specifies a 'Platform', then you have to add this Condition. I would imagine that just specifying a RuntimeIdentifier would imply a Platform, but changing the build to use RuntimeIdentifier and not Platform is probably out of scope.

mschofie (approved):

PR #596, opened 2026-02-21
Add multi-headed MSIX package sample apps

guimafelipe:
## Summary
- Adds two sample apps under `Samples/MultiHeadedPackage/` demonstrating multiple `<Application>` elements in a single package manifest
- **cs-winui-msix**: WinUI 3 primary app + console secondary app packaged together in one MSIX via single-project packaging
- **cs-wpf-sparse**: WPF primary app (with register/unregister/restart UI) + WPF secondary app in one sparse package

These samples support the multi-headed MSIX packaging feature added in WindowsAppSDK (microsoft/WindowsAppSDK#5586).

## Description

Multi-headed packages contain multiple `<Application>` entries in a single `Package.appxmanifest`. Each application gets its own Start menu tile and can be launched independently, while sharing the same package identity and installation lifecycle.

The **MSIX sample** uses single-project MSIX packaging. The primary WinUI app owns the manifest and uses `$targetnametoken$` tokens. The secondary console app is included via `<ProjectReference>` with its executable hardcoded in the manifest as `SecondaryApp.exe` with `EntryPoint="Windows.FullTrustApplication"`. After deployment, both apps appear as separate Start menu entries.

The **sparse sample** demonstrates the same multi-headed pattern with runtime-registered sparse packages. The primary WPF app provides Register/Unregister/Restart buttons using `PackageManager.AddPackageByUriAsync`. The secondary WPF app uses `GetCurrentPackageFullName` to detect whether it has package identity, which it will once the sparse package is registered.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Samples/MultiHeadedPackage/README.md
```diff
@@ -0,0 +1,65 @@
...
+description: "Shows how to package multiple executables in a single MSIX or sparse package"
+extendedZipContent:
+- path: LICENSE
+  target: LICENSE
+---
+# Multi-headed MSIX package sample
+
+These samples demonstrate how to create **multi-headed packages** — MSIX or sparse packages that contain multiple `<Application>` elements in a single `Package.appxmanifest`. Each application entry gets its own Start menu tile and can be launched independently, while sharing the same package identity and installation lifecycle.
+
+## Samples
```
Scottj1s, line 19: nice to have these samples.  the most interesting scenario, which I think would also be the most common, is missing: two winui apps.  I suspect they would be 'oversharing' assets, resources.pris, etc.  can we add a sample that would expose such collisions and work around by locating assets in separate msix folders?

guimafelipe: This is one of the samples that I will need to craft with more care. I can't just use the templates, as there will be a lot of name conflicts. I am thinking of starting with these two samples first.

What do you mean by putting assets in separate msix folders?

Scottj1s: I spent a couple hours trying to get a multi-headed wapproj working.  Turns out it doesn't just work out of the box, via two ProjectReferences added in the IDE.  

Others have reported the same experience:
https://github.com/microsoft/WindowsAppSDK/issues/1650
https://github.com/microsoft/WindowsAppSDK/issues/4583
https://github.com/microsoft/microsoft-ui-xaml/issues/10982

Multiple suggestions from copilot did not help:
```
vcxproj:
    <AppxPackageDir>SecondApp\</AppxPackageDir>
wapproj:
        <ProjectReference Include="..\SecondApp\SecondApp.vcxproj">
            <PackageFolder>SecondApp</PackageFolder>
            <Properties>PackLayout=false</Properties>
        </ProjectReference>
appxmanifest.xml:
    <Application Executable="SecondApp\SecondApp.exe"
AppxLayoutMap.xml:
    <FileMapping SourcePath="...\SecondApp\*" DestinationPath="SecondApp\" />
etc...

```
The only solution I could find was suggested in an issue above - manually eliminate conflicts with unique naming.  That's an awful workaround - we should strive to automatically support any two apps, segregating their content in separate MSIX folders.   The Xaml compiler may need to be enlightened to support loose Xbf files (in debug builds) with custom TargetPaths.

guimafelipe: Yes, I had the same experience trying to make it work on WAP. It is not a great experience. I tried also so spin up some agents to keep trying different solutions for that and most results were overengineered stuff.

My idea on how to tackle this was to do the "manually eliminate conflicts with unique naming" approach, but trying to make it automatically in the build system.

I am trying to understand the implications of generating two msix folders. Would they still be inside the same package? We can talk more about that offline too

PR #606, opened 2026-03-16
samples: sync NotesApp with latest ACI API changes

surajkumar-08:
## Summary

### Namespace & API renames
 - Update namespace from `Microsoft.Windows.AI.Search.Experimental.AppContentIndex` to `Microsoft.Windows.Search.AppContentIndex`
 - Rename `Remove` → `RemoveContentItem`, `RemoveAll` → `RemoveAllContentItems`
 - Use `RegionOfInterest` (on `AppManagedImageQueryMatch`) instead of `Subregion` (which remains on `AppManagedOcrTextQueryMatch` for OCR matches)
 
 ### New API usage
 - Add pre-flight system capability checks via `GetIndexCapabilitiesOfCurrentSystem()` before creating an index
 - Subscribe to `AppContentIndexListener` events (`IndexCapabilitiesChanged`, `IndexStatisticsChanged`, `ContentItemStatusChanged`) for live index monitoring
 - Drive search UI progress bar from `IndexStatisticsChanged` events
 - Log index statistics via `GetIndexStatistics()` on startup
 
 ### Resource management & robustness
 - Dispose `AppContentIndexer` in `MainWindow.Closed` handler
 - Add `IDisposable` to `SearchViewModel`; call `Dispose()` via `SearchView.Unloaded` to release `CancellationTokenSource`
 - Replace `int.Parse` with `int.TryParse` for image match `ContentId` to avoid `FormatException`
 - Replace `throw new Exception` with `throw new InvalidOperationException`
 
 ### Cleanup
 - Update WinAppSDK from 2.0.0-experimental3 to 2.0.0-preview1
 - Remove stale commented-out internal project references from csproj
 - Fix README: correct solution filename, update SDK version reference
 - Clean up `Package.appxmanifest` identity for public sample

## Checklist

### DIFF
--- a/Samples/AppContentSearch/cs-winui/Pages/SettingsPage.xaml
+++ b/Samples/AppContentSearch/cs-winui/Pages/SettingsPage.xaml
@@ -2,8 +2,7 @@
     x:Class="Notes.Pages.SettingsPage"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:local="using:Notes.Pages"
-    xmlns:toolkit="using:CommunityToolkit.WinUI.Controls">
+    xmlns:local="using:Notes.Pages">
 
     <Page.Resources>
         <x:Double x:Key="SettingsCardSpacing">4</x:Double>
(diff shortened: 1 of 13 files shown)


surajkumar-08:
@abhmondal_microsoft can you please provide your review.

surajkumar-08:
@microsoft-github-policy-service agree company="Microsoft"

Arvind2311:
### Minor: `_searchText` thread safety

In `SearchViewModel`, `_searchText` is written on the UI thread (via `HandleTextChanged`) and read from `ResultChanged` callbacks (which may fire on a background thread).

The code does correctly snapshot the value at the top of `RefreshQuerySessionResultsAsync`:
```csharp
string searchText = _searchText;  // snapshot
```

But then later checks staleness against the live field:
```csharp
!string.Equals(searchText, _searchText, StringComparison.Ordinal)
```

This comparison is inherently racy — `_searchText` could be torn or stale depending on timing. In practice this is unlikely to cause visible issues in a sample app since the worst case is an extra redundant UI update, but it's worth being aware of.

A simple mitigation would be marking `_searchText` as `volatile`, or using `Interlocked.Exchange`/`Interlocked.CompareExchange` with a reference. Not blocking, just a heads-up.

Samples/AppContentSearch/cs-winui/Utils/Utils.cs
```diff
@@ -206,8 +229,71 @@ await Task.Run(() =>
...
+            }
+
+            var note = await context.Notes.FindAsync(attachment.NoteId);
+            if (note == null)
+            {
+                Debug.WriteLine($"Skipping OCR match; note not found for attachment ID: {attachmentId}");
+                return null;
+            }
+
+            string fragment = match.GetType().GetProperty("Fragment")?.GetValue(match) as string ?? string.Empty;
```
Arvind2311, line 272: The `GetType().GetProperty("Fragment")` pattern for reading OCR results is the only reflection-based property access in the entire samples repo. If the API renames that property in a future preview, this silently returns `null` and displays empty text with no error.

Is this a temporary workaround for a property not yet projected in the preview SDK? If so, could we add a `// TODO: Replace with direct property access when projected` comment so developers copying this sample know it's not the recommended pattern?

Samples/AppContentSearch/cs-winui/ViewModels/SearchViewModel.cs
```diff
@@ -74,54 +160,172 @@ private async Task Search()
...
+            catch (Exception ex)
+            {
+                Debug.WriteLine($"Search failed: {ex.Message}");
+                // Handle other exceptions as needed
+            }
+        }
+
+        private void TextQuerySession_ResultChanged(AppIndexTextQuerySession sender, object args)
+        {
+            _ = RefreshQuerySessionResultsAsync();
```
Arvind2311: `_ = RefreshQuerySessionResultsAsync()` discards the Task, so if the query session throws (e.g., after disposal or on an unsupported system), the exception vanishes and search silently stops working. Since this sample teaches developers how to use the Content Search API, would it be worth wrapping this in a `try/catch` with a logged or displayed error? Even a `Debug.WriteLine` would make failures visible when developers are adapting this code.

Samples/AppContentSearch/cs-winui/Pages/SettingsPage.xaml
```diff
@@ -29,27 +28,35 @@
...
+                                <ComboBoxItem x:Name="PhiSilicaComboBoxItem" Content="Windows AI APIs LanguageModel (Phi Silica)" IsSelected="True"/>
+                                <ComboBoxItem x:Name="AzureComboBoxItem" Content="Azure OpenAI"/>
+                                <ComboBoxItem x:Name="FoundryComboBoxItem" Content="Foundry Local"/>
+                            </ComboBox>
 
-                        <toolkit:SettingsExpander.Items>
                             <!-- Azure Settings Card -->
-                            <toolkit:SettingsCard x:Name="AzureSettingsCard" Header="AzureSettings" Visibility="Collapsed" HorizontalContentAlignment="Left" ContentAlignment="Left">
-                                <StackPanel Margin="-42,-24,0,12" Spacing="8">
+                            <Border x:Name="AzureSettingsCard"
```
Arvind2311, line 44: The switch from `SettingsCard` to `Border` removes the keyboard navigation and Narrator landmarks that `SettingsCard` provides out of the box. Since developers often copy sample XAML patterns directly, could we preserve accessibility here — either by keeping `SettingsCard` or adding `AutomationProperties` to the replacement controls?

Samples/AppContentSearch/cs-winui/ViewModels/SearchViewModel.cs
surajkumar-08: Done

Samples/AppContentSearch/cs-winui/Pages/SettingsPage.xaml
surajkumar-08: Done

Samples/AppContentSearch/cs-winui/Utils/Utils.cs
surajkumar-08: Done

PR #619, opened 2026-04-14
Fix WPF inference UI thread blocking and add busy indicator

yeelam-gordon:
## Summary

The WPF WindowsML sample's `RunInference` call was blocking the UI thread, risking freezes on larger models. This PR fixes the threading issue and adds a visual busy indicator.

### 1. Move inference to background thread (`Task.Run`)
- `InferenceEngine.RunInference` is synchronous (calls `session.Run`). Even though the calling method is `async`, prior `await` calls resume on the UI thread via WPF's `SynchronizationContext`, so `RunInference` still blocks the UI thread.
- Wrapped in `await Task.Run(...)` to offload to a thread pool thread.

### 2. Add indeterminate ProgressBar busy indicator
- Added a `ProgressBar` with `IsIndeterminate=True` that appears during inference and model loading.
- Added `SetBusy(bool)` helper that toggles the progress bar and disables interactive buttons (Run Inference, Select Image, Load/Reload Model) to prevent double-clicks.
- Applied to both `RunInferenceButton_Click` and `ReloadSessionButton_Click` with `try/finally`.
- Removed the old `Dispatcher.Invoke` render-flush workaround (no longer needed).

## Testing
- Built successfully with `dotnet build` (0 errors, 0 warnings)
- Ran the app via `dotnet run` and verified the window launches correctly

Fixes ADO #61791035

### DIFF
--- a/Samples/WindowsML/cs-wpf/MainWindow.xaml
+++ b/Samples/WindowsML/cs-wpf/MainWindow.xaml
@@ -62,6 +62,10 @@
             </Grid>
         </Grid>
 
+        <!-- Busy indicator (indeterminate progress bar, hidden by default) -->
+        <ProgressBar x:Name="BusyIndicator" Grid.Row="0" IsIndeterminate="True" Visibility="Collapsed"
+                     Height="3" VerticalAlignment="Bottom" Margin="0,0,0,-5"/>
+
         <!-- Image Display Row -->
         <GroupBox Header="Selected Image" Grid.Row="1" Margin="0,0,0,10" MinHeight="250">
             <Border BorderBrush="Gray" BorderThickness="1" Background="LightGray">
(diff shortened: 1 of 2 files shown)


nieubank:
Thanks for putting this together. During triage for ADO 61791035, we confirmed this is a WPF sample responsiveness/UX issue rather than a WinML API correctness issue or a sample-compilation blocker. The current SqueezeNet sample can still demonstrate WinML functionality; the proposed change adds busy-state behavior and moves work off the UI thread, which is useful polish but higher churn than we want for this bug bash pass.

Recommendation: close/defer this PR for ADO 61791035. If the team wants a broader WPF UX polish pass, it should be tracked separately with the sample convergence work rather than ported as an isolated WinML bug fix.

PR #620, opened 2026-04-14
Fix --ep_policy DISABLE: handle gracefully instead of crashing

yeelam-gordon:
## Summary

Fixes AB#61790993 — \\--ep_policy DISABLE\\ fails validation and crashes downstream code.

## Problem

\\DISABLE\\ means *do not call \\SetEpSelectionPolicy\\* — use ONNX Runtime built-in defaults (CPU + DML). Since \\OrtExecutionProviderDevicePolicy\\ has no \\DISABLE\\ enum value, the parser correctly leaves \\p_policy\\ unset. However, three layers then fail:

1. **ArgumentParser** validation requires \\p_policy\\ or \\p_name\\ to be set
2. **InferenceEngine.cpp** falls to else branch → calls \\ConfigureSelectedExecutionProvider\\ with empty ep_name → error
3. **ModelManager.cs** \\CreateSession\\ and \\ResolveActualModelPath\\ throw \\Exception\\ when neither is set

## Fix

- **ArgumentParser (C++/C#)**: Track DISABLE via a local flag, bypass the *must specify ep_policy or ep_name* validation
- **InferenceEngine.cpp**: Three-way branch — ep_policy → explicit EP → defaults (DISABLE)
- **ModelManager.cs**: \\CreateSession\\ and \\ResolveActualModelPath\\ gracefully fall through to ONNX Runtime defaults instead of throwing

## Semantics

\\DISABLE\\ ≠ \\DEFAULT\\:
- **DEFAULT**: actively calls \\SetEpSelectionPolicy(DEFAULT)\\ → ONNX Runtime automatic EP selection
- **DISABLE**: skips EP configuration entirely → falls back to built-in CPU + DML

## Testing

Verified with \\CSharpConsoleDesktop\\ sample:
- \\--ep_policy DISABLE\\ → runs inference successfully with ONNX Runtime defaults ✅
- \\--ep_policy DEFAULT\\ → uses EP selection policy as before ✅
- No args → validation error fires correctly ✅

### DIFF
--- a/Samples/WindowsML/Shared/cpp/ArgumentParser.cpp
+++ b/Samples/WindowsML/Shared/cpp/ArgumentParser.cpp
@@ -20,6 +20,8 @@ namespace Shared
             return false;
         }
 
+    bool disable_ep = false; // tracks --ep_policy DISABLE locally for validation
+
     for (size_t i = 1; i < arguments.size(); ++i)
     {
         if (arguments[i] == L"--compile")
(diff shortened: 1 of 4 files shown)


yeelam-gordon:
Closed as another PR fix this: https://github.com/microsoft/WindowsAppSDK-Samples/pull/635

PR #622, opened 2026-04-14
Auto-load model and run inference on WinUI app launch

yeelam-gordon:
Instead of requiring the user to manually select an execution provider and click **Load / Reload Model**, the WinUI WindowsML sample now automatically loads the model with the default EP and runs inference on startup.

If auto-load fails (e.g., no GPU available for DML), the error is shown and the user can still manually select a different EP and retry.

## Changes

- **Samples/WindowsML/cs-winui/MainWindow.xaml.cs** — After EP initialization, automatically calls LoadModelAsync() and LoadAndRunDemoAsync() with graceful error handling fallback.

## Testing

- Built and deployed on a Hyper-V VM (no GPU) — app auto-runs with CPU EP successfully
- Manual reload button still works for switching EPs

### DIFF
--- a/Samples/WindowsML/cs-winui/MainWindow.xaml.cs
+++ b/Samples/WindowsML/cs-winui/MainWindow.xaml.cs
@@ -78,8 +78,24 @@ namespace WindowsMLSample
                 EpCombo.SelectionChanged += EpCombo_SelectionChanged;
 
                 providerInfo.AppendLine("========================================");
-                providerInfo.AppendLine("Select an execution provider (and device if required) then click 'Load / Reload Model'.");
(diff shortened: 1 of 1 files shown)


nieubank:
Thanks for putting this together. During triage for ADO 61791037, I could not reproduce the reported issue as filed in the current `windows-ml` sample copy: `ReloadModelButton` is wired to `ReloadModelButton_Click`, and that handler loads the model and runs demo inference.

This PR changes the sample behavior to auto-load and run inference on startup, which looks like a UX enhancement rather than a bug fix for an unwired demo path.

Recommendation: close this PR for ADO 61791037. If the team wants auto-load-on-launch behavior, I recommend tracking it as a separate sample UX enhancement.

PR #623, opened 2026-04-14
Implement GetTestMessage in CppConsoleDll using Ort::Env initialization

yeelam-gordon:
## Summary

The CppConsoleDll README documented a \GetTestMessage\ function that was never implemented. The function typedef existed in the client code from day one but was never wired up.

## Changes

- **WindowsMLWrapper.cpp**: Added \GetTestMessage()\ which creates an \Ort::Env\ — the canonical ORT initialization step. If it succeeds, the ONNX Runtime is genuinely loaded and functional, returning: \WindowsML DLL is working correctly!\. Includes \Ort::Exception\ catch for specific error reporting.
- **WindowsMLWrapper.h**: Added \GetTestMessage()\ declaration with doc comment.
- **ConsoleClient.cpp**: Wired up \GetTestMessage\ call, added \Test message:\ prefix to match README expected output, restored null-handling for both functions.

## Verified Output

\\\
Calling GetTestMessage from DLL...
Test message: WindowsML DLL is working correctly!
Calling GetOrtVersionString from DLL...
ONNX Runtime version: 1.23.4
DLL functions called successfully!
\\\

This matches the README expected output exactly.

## Bug

Fixes #61791036

### DIFF
--- a/Samples/WindowsML/cpp/CppConsoleDll/ConsoleClient/ConsoleClient.cpp
+++ b/Samples/WindowsML/cpp/CppConsoleDll/ConsoleClient/ConsoleClient.cpp
@@ -5,8 +5,8 @@
 #include <iostream>
 
 // Function pointer types for the DLL exports
-typedef char*(__stdcall* GetOrtVersionStringFunc)();
-typedef char*(__stdcall* GetTestMessageFunc)();
+typedef char*(*GetOrtVersionStringFunc)();
+typedef char*(*GetTestMessageFunc)();
 
 int main()
 {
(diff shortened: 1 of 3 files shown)


nieubank:
Thanks for catching this mismatch. I opened a smaller replacement PR here: https://github.com/microsoft/WindowsAppSDK-Samples/pull/641

I think the better fix is to align the README with what the sample actually exports/calls, rather than adding a new GetTestMessage DLL export just to satisfy the README. The replacement PR removes the unsupported GetTestMessage expected-output lines and deletes the unused GetTestMessageFunc typedef, with no new sample API surface or error paths.

Recommendation: close this PR in favor of #641.

nieubank:
Closed in favor of #641

PR #625, opened 2026-04-14
Add missing App.InitializeComponent() call in WinUI sample

yeelam-gordon:
Fixes ADO #61791038

The WinUI cs-winui sample was missing the App.InitializeComponent() call, which could cause XAML resources to not load properly.

### DIFF
--- a/Samples/WindowsML/cs-winui/App.xaml.cs
+++ b/Samples/WindowsML/cs-winui/App.xaml.cs
@@ -11,6 +11,11 @@ namespace WindowsMLSample
 {
     public partial class App : Application
     {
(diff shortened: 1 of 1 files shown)


nieubank:
Reviewed for ADO 61791038. This PR is the right minimal fix: `App.xaml` defines application resources, and `App.xaml.cs` currently has no `App()` constructor calling `InitializeComponent()`, so those resources may not be loaded. The one-file change here matches the bug and is low risk.

Recommendation: merge this PR and resolve ADO 61791038 as Fixed after merge.

nieubank (approved):

PR #626, opened 2026-04-14
Wire up model variant selection instead of hardcoded SqueezeNet path

yeelam-gordon:
Fixes ADO #61791040

The cs-wpf, cs-winui, and cs-winforms samples had hardcoded model paths ignoring the variant selection UI. This wires up the combo box selection to actually load the chosen model variant.

### DIFF
--- a/Samples/WindowsML/cs-winforms/MainForm.cs
+++ b/Samples/WindowsML/cs-winforms/MainForm.cs
@@ -126,7 +126,10 @@ namespace WindowsMLWinFormsSample
 
             var options = new Options
             {
-                ModelPath = "SqueezeNet.onnx",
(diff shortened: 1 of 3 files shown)


nieubank:
Thanks for putting this together. During triage for ADO 61791040, we could not reproduce the reported issue as a broken hardcoded model path in the current samples: the WinUI/WPF/WinForms samples use the app-local relative model name `SqueezeNet.onnx`, and the shared/console paths support custom model paths via `--model`.

This PR adds model-variant selection behavior, which looks like a sample UX enhancement rather than a fix for the reported hardcoded-path issue.

Recommendation: close/defer this PR for ADO 61791040. If the team wants model-picker or model-variant UX work, it should be tracked separately from this bug bash item.

PR #634, opened 2026-04-30
Bug Bash - EP+Performance compile paths and fixes

mahabayana:
## Changes
 Bug 1 – Missing ready state output after EnsureReadyAsync

   - No feedback was printed after calling EnsureReadyAsync(), making EP readiness hard to debug
   - Added a Console.WriteLine to show the updated ReadyState after the call completes

  Bug 2 – Remove DISABLE from ep_policy options

   - DISABLE was listed as a valid --ep_policy value but just set policy to null
   - Unknown values silently fell back to DISABLE instead of reporting an error
   - Removed DISABLE, unknown values now throw an error with valid options listed
   - Updated help text and READMEs in both C++ and C#

  Bug 3 – EP+Performance-specific compiled model paths

   - Compiled models always saved as SqueezeNet_ctx.onnx, so different EP/perf configs overwrote each other
   - Added GenerateCompiledModelPath() in both C++ and C# that encodes EP policy/name, device type, and perf mode into the filename
   - Example: SqueezeNet_ctx_PREFER_NPU_MaxPerformance.onnx
   - Users can still override with --compiled_output
   - Updated READMEs with examples of auto-generated filenames

### DIFF
--- a/Samples/WindowsML/Shared/cpp/ModelManager.cpp
+++ b/Samples/WindowsML/Shared/cpp/ModelManager.cpp
@@ -256,8 +256,7 @@ namespace Shared
     std::filesystem::path ModelManager::GenerateCompiledModelPath(
         const std::filesystem::path& modelPath,
         const std::filesystem::path& executableFolder,
-        const CommandLineOptions& options,
-        Ort::Env& env)
+        const CommandLineOptions& options)
     {
         // If user explicitly specified --compiled_output, use it as-is
         if (!options.output_path.empty())
--- a/Samples/WindowsML/Shared/cpp/ModelManager.cpp
+++ b/Samples/WindowsML/Shared/cpp/ModelManager.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE.md in the repo root for license information.
 #include "ModelManager.h"
 #include <iostream>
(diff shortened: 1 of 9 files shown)


Samples/WindowsML/Shared/cpp/ModelManager.cpp
```diff
@@ -253,5 +253,57 @@ namespace Shared
         return ModelVariant::Default;
     }
 
+    std::filesystem::path ModelManager::GenerateCompiledModelPath(
+        const std::filesystem::path& modelPath,
+        const std::filesystem::path& executableFolder,
+        const CommandLineOptions& options,
+        Ort::Env& env)
```
Kevin-Taha: Is env needed? it's never used.

Samples/WindowsML/Shared/cs/ModelManager.cs
```diff
@@ -110,6 +110,60 @@ public static string GetModelVariantPath(string executableFolder, ModelVariant v
...
+
+            string fileName = $"{baseName}_ctx{suffix}.onnx";
+            Console.WriteLine($"Compiled model path: {Path.Combine(executableFolder, fileName)}");
+            return Path.Combine(executableFolder, fileName);
+        }
+
+        /// <summary>
+        /// Build a device-identifying suffix for the compiled model filename.
+        /// </summary>
+        private static string BuildDeviceSuffix(Options options, OrtEnv? ortEnv)
```
Kevin-Taha: same here, unused ortEnv

Samples/WindowsML/cpp/CppConsoleDesktop/README.md
```diff
@@ -11,11 +11,11 @@ This sample demonstrates how to use ONNX Runtime in a C++ desktop application, f
 ```
 CppConsoleDesktop.exe [options]
 Options:
-  --ep_policy <policy>  Set execution provider policy (NPU, CPU, GPU, DEFAULT, DISABLE). Default: DISABLE
+  --ep_policy <policy>  Set execution provider policy (NPU, CPU, GPU, DEFAULT)
```
adrastogi: I see that there's an open PR for addressing the issue w/ the DISABLE option: (#620). How would you all like to reconcile this? @mahabayana @yeelam-gordon

mahabayana: @adrastogi  @yeelam-gordon  
The samples require either --ep_policy or --ep_name to be set. The following scenarios fail due to DISABLE:

  Scenario 1 : User explicitly passes DISABLE: The user runs --ep_policy DISABLE. The parser sets ep_policy = null and ep_name remains empty. Validation sees neither is set and throws: "Missing EP selection" - even though the user did specify a policy.

  Scenario 2 : User passes an invalid value (e.g., a typo): The user runs --ep_policy NPUU. The parser prints "Unknown EP policy: NPUU, using default (DISABLE)" and silently sets ep_policy = null. Validation again throws "Missing EP selection" : masking the real error (an invalid value)

 Thus I would propose this resolution

   - Removed DISABLE as a valid --ep_policy option. If the user's intention is to default to CPU behavior, they can use DEFAULT instead, making DISABLE redundant.
   - Unknown values now throw immediately with the list of valid options (NPU, CPU, GPU, DEFAULT), so typos and invalid   inputs are caught at parse time rather than falling through to a incorrect validation error

adrastogi (approved):

mahabayana:
@microsoft-github-policy-service agree company="Microsoft"

PR #657, opened 2026-08-10
Build C++ samples with C++20 to fix STL1011, and map VS 2026 to v145

iamAbhi-916:
## Description

Fixes the C++ samples failing to build on current toolsets: `error STL1011` from the deprecated `/await` switch, `error MSB8020` on Visual Studio 2026, and the template conformance errors in `UXFrameworksOnIslands`.

Fixes ADO 62276972.

## Root cause

**STL1011** — `Microsoft.Windows.CppWinRT.props` defaults `ClCompile.LanguageStandard` to `stdcpp17` when a project does not set one, and `Microsoft.Windows.CppWinRT.targets` then adds `/await` for `stdcpp17`. Under `/await`, `winrt/base.h` takes the experimental branch and includes the `experimental/coroutine` header, which now hard-errors. This hits every C++/WinRT sample, whether or not it uses coroutines.

**MSB8020** — `Build.Common.Cpp.props` has no `BuildToolVersion` entry for VS 2026, so `PlatformToolset` is assigned an empty value and MSBuild falls back to `v100`.

**C2061 / C4864** — `UXFrameworksOnIslands` relies on unqualified lookup finding members of a dependent base, which current MSVC rejects under two-phase lookup.

## Fix

Set `LanguageStandard` to `stdcpp20` in `Samples/Directory.Build.props`, and map `BuildToolVersion` 18 to `v145` in `Samples/Build.Common.Cpp.props`.

**Why 8 files for the language standard:** `Directory.Build.props` is imported before the project body, so the 7 projects that set their own `stdcpp17` overwrite it — they would still get `/await` and still fail STL1011, so they are updated individually. The other 33 set no value of their own, so they inherit it and need no edit.

`SampleWidgetProviderApp` was the only C++ sample that neither imported `Build.Common.Cpp.props` nor set its own `PlatformToolset`, so it now imports the shared props like the other C++ samples.

For `UXFrameworksOnIslands`: `TextVisual<T>` derives from `D2DSprite<T>`, so `ContainerVisual`, `GetVisual`, `GetBackgroundColor` and `InvalidateContent` are introduced with using-declarations, which also resolves the cascading `C2065`/`C3861`/`C2672` errors in `TextVisual.cpp` and in the `InsertTextVisual` callers. In `AutomationFragment::RemoveChild` the lambda parameter is dependent, so the member template call needs the `template` disambiguator. These are pre-existing conformance gaps unrelated to the language standard — they reproduce at `stdcpp17` with STL1011 suppressed.

## Validation

Built **all 41 C++/WinRT sample solutions** at x64 Release on Visual Studio 2026 (MSVC 14.51.36231), before and after the change, with no command-line overrides:

| | Built | Failed |
|---|---|---|
| `main` | 15 | 26 |
| this PR | 32 | 9 |

Runtime: the rebuilt unpackaged console samples launch without crashing and reach the Windows App SDK bootstrapper. A full functional run was not possible on the validation machine because the matching Windows App Runtime is not installed there.

## Target Release

1.8

## Checklist

- [x] **[For new samples]**: Samples have completed the [sample guidelines checklist](https://github.com/microsoft/WindowsAppSDK-Samples/blob/main/docs/samples-guidelines.md#checklist) and follow [standardization/naming guidelines](https://github.com/microsoft/WindowsAppSDK-Samples/blob/main/docs/samples-guidelines.md#standardization-and-naming).
- [x] Microsoft employees only: you can validate your changes by following the instructions here: https://www.osgwiki.com/wiki/WindowsAppSDK-Samples

Builds were run at x64 Release; x86/ARM64 and Debug are unverified. The change is a language-standard switch applied uniformly across all configurations, so it is not configuration-specific.

### DIFF
--- a/Samples/AppLifecycle/Activation/cpp/cpp-win32-packaged/CppWinMainActivation/CppWinMainActivation.vcxproj
+++ b/Samples/AppLifecycle/Activation/cpp/cpp-win32-packaged/CppWinMainActivation/CppWinMainActivation.vcxproj
@@ -129,7 +129,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>WIN32;_DEBUG;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
@@ -144,7 +144,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>WIN32;NDEBUG;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
@@ -159,7 +159,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
@@ -172,7 +172,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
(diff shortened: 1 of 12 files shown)


Samples/Directory.Build.props
```diff
@@ -7,6 +7,13 @@
 
   <ItemDefinitionGroup>
     <ClCompile>
+      <!-- Microsoft.Windows.CppWinRT.props defaults LanguageStandard to stdcpp17 when a project
```
godlytalias: This comment can be removed i think

iamAbhi-916: okay , will clean it up

Samples/AppLifecycle/Activation/cpp/cpp-win32-packaged/CppWinMainActivation/CppWinMainActivation.vcxproj
```diff
@@ -129,7 +129,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>WIN32;_DEBUG;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
```
godlytalias, line 132: Can you handle for https://github.com/microsoft/WindowsAppSDK-Samples/tree/main/Samples/Islands/UXFrameworksOnIslands as well in this PR?

task.ms/62276972

iamAbhi-916: yeah it would be better to add this too, will update

Samples/Directory.Build.props
iamAbhi-916: done

Samples/AppLifecycle/Activation/cpp/cpp-win32-packaged/CppWinMainActivation/CppWinMainActivation.vcxproj
iamAbhi-916: done

godlytalias (approved):

HariniMalothu17 (approved):

PR #671, opened 2026-09-09
Windows ML: Add compiled model compatibility samples

adrastogi:
## Description

Updates the Windows ML samples to demonstrate how to use ONNX Runtime's compiled-model compatibility APIs before reusing a cached compiled model.

The C++, C#, and Python sample implementations now:

- Extract execution-provider-specific compatibility metadata from cached models.
- Group candidate devices by execution provider and validate each group with `GetModelCompatibilityForEpDevices`.
- Reuse a cached model only when compatibility is reported as `EP_SUPPORTED_OPTIMAL`.
- Reject caches with missing, incompatible, or non-optimal metadata and either recompile when requested or fall back to the original ONNX model.
- Write compiled models to temporary output and replace the cache only after successful compilation.
- Embed execution-provider context in the compiled model so the cache does not depend on sidecar files.

The associated documentation now explains the compatibility statuses, same-EP device-group requirement, cache-reuse policy, and fallback behavior.

## Target Release

2.3

## Checklist

### DIFF
--- a/Samples/WindowsML/Directory.Packages.props
+++ b/Samples/WindowsML/Directory.Packages.props
@@ -5,7 +5,9 @@
   <ItemGroup>
     <!-- WindowsAppSDK packages -->
     <PackageVersion Include="Microsoft.ML.OnnxRuntimeGenAI.WinML" Version="0.14.0" />
-    <PackageVersion Include="Microsoft.Windows.AI.MachineLearning" Version="2.1.1" />
+    <!-- Managed compatibility metadata extraction requires the newer Windows ML binding.
+         Native samples remain on packages.config 2.1.1, whose C++ API already includes it. -->
+    <PackageVersion Include="Microsoft.Windows.AI.MachineLearning" Version="2.3.42" />
     <PackageVersion Include="Microsoft.WindowsAppSDK" Version="2.1.3" />
     <PackageVersion Include="Microsoft.WindowsAppSDK.AI" Version="2.1.10" />
     <PackageVersion Include="Microsoft.WindowsAppSDK.Base" Version="2.0.4" />
--- a/Samples/WindowsML/Shared/cpp/ModelManager.cpp
+++ b/Samples/WindowsML/Shared/cpp/ModelManager.cpp
@@ -34,6 +34,30 @@ namespace Shared
         std::cout << "Compiling model from " << modelPath << std::endl;
         std::cout << "Output path: " << compiledModelPath << std::endl;
 
+        std::error_code pathError;
+        const bool compiledModelExists = std::filesystem::exists(compiledModelPath, pathError);
+        if (pathError)
+        {
+            const std::string message =
+                "Failed to inspect compiled model output path: " + pathError.message();
+            std::cerr << message << std::endl;
+            return ortApi.CreateStatus(ORT_FAIL, message.c_str());
+        }
+
+        if (compiledModelExists && std::filesystem::equivalent(modelPath, compiledModelPath, pathError))
+        {
+            const std::string message = "Compilation output path must be different from the original model path.";
+            std::cerr << message << std::endl;
+            return ortApi.CreateStatus(ORT_INVALID_ARGUMENT, message.c_str());
+        }
+        if (pathError)
+        {
+            const std::string message =
+                "Failed to compare input and compiled model paths: " + pathError.message();
+            std::cerr << message << std::endl;
+            return ortApi.CreateStatus(ORT_FAIL, message.c_str());
+        }
+
         // Get compile API
         const OrtCompileApi* compileApi = ortApi.GetCompileApi();
         if (!compileApi)
--- a/Samples/WindowsML/Shared/cpp/ModelManager.cpp
+++ b/Samples/WindowsML/Shared/cpp/ModelManager.cpp
@@ -1,10 +1,23 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE.md in the repo root for license information.
 #include "ModelManager.h"
-#include <iostream>
-#include <fstream>
 #include <chrono>
+#include <fstream>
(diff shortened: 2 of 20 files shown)


Samples/WindowsML/Shared/cpp/ModelManager.cpp
```diff
@@ -26,9 +39,13 @@ namespace Shared
         if (!compileApi)
         {
             std::cerr << "Failed to get compile API" << std::endl;
-            return nullptr;
+            return ortApi.CreateStatus(ORT_FAIL, "Failed to get compile API");
         }
 
+        const std::filesystem::path temporaryModelPath = CreateTemporaryCompiledModelPath(compiledModelPath);
```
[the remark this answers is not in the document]
adrastogi: Fixed in `53c5aff0` by rejecting compilation when the requested native output path refers to the existing input model. Verified that identical paths leave the input unchanged and that distinct-output compilation still succeeds.

Samples/WindowsML/Shared/cs/ModelManager.cs
```diff
@@ -401,33 +510,175 @@ public static string ResolveActualModelPath(Options options, string modelPath, s
...
+                    candidate => candidate.HardwareDevice.Type.ToString().Equals(
+                        options.DeviceType,
+                        StringComparison.OrdinalIgnoreCase));
+                return device == null ? [] : [(group.EpName, [device])];
+            }
+
+            HashSet<string> policyDeviceTypes = GetPolicyDeviceTypes(
+                options.EpPolicy ?? ExecutionProviderDevicePolicy.DEFAULT);
+
+            // Preserve EP grouping while including the policy's preferred hardware and CPU fallback.
```
nieubank: Agentic review: [Blocking] [Intent: Defect] [Severity: High] [Area: Correctness]

**Finding:** Policy-mode compatibility checks every EP that exposes the preferred device type or CPU, then skips groups with no metadata. ORT's policy selector picks specific ordered devices, so another EP's EP_SUPPORTED_OPTIMAL metadata can approve a cache that was never validated for the EP that will actually run.

**Recommendation:** Validate the devices SetEpSelectionPolicy will select (C++ SelectDeviceGroups / Python _filter_ep_devices_by_types have the same aggregation).

Samples/WindowsML/Directory.Packages.props
```diff
@@ -5,7 +5,9 @@
   <ItemGroup>
     <!-- WindowsAppSDK packages -->
     <PackageVersion Include="Microsoft.ML.OnnxRuntimeGenAI.WinML" Version="0.14.0" />
-    <PackageVersion Include="Microsoft.Windows.AI.MachineLearning" Version="2.1.1" />
+    <!-- Managed compatibility metadata extraction requires the newer Windows ML binding.
+         Native samples remain on packages.config 2.1.1, whose C++ API already includes it. -->
+    <PackageVersion Include="Microsoft.Windows.AI.MachineLearning" Version="2.3.42" />
```
nieubank, line 9: Agentic review: [Blocking] [Intent: Defect] [Severity: High] [Area: Packaging]

**Finding:** Microsoft.Windows.AI.MachineLearning 2.3.42 is not on WinAppSDK-SampleDeps (index: 2.1.74, 2.1.1; nupkg 404). The PR also adds a localpackages source that is not on main. A clean public clone cannot restore the four C# samples.

**Recommendation:** Publish 2.3.42 to the sample feed (or pin a published version) and drop the localpackages source unless it is an optional gitignored overlay.

Samples/WindowsML/cpp/CppConsoleDesktop/README.md
```diff
@@ -94,6 +110,7 @@ compileApi->CreateModelCompilationOptionsFromSessionOptions(env, sessionOptions,
 // Set input and output paths
 compileApi->ModelCompilationOptions_SetInputModelPath(compileOptions, modelPath.c_str());
 compileApi->ModelCompilationOptions_SetOutputModelPath(compileOptions, compiledModelPath.c_str());
+compileApi->ModelCompilationOptions_SetEpContextEmbedMode(compileOptions, true);
```
nieubank: Agentic review: [Non-blocking] [Intent: Defect] [Severity: Medium] [Area: Documentation]

**Finding:** Prose describes temp compile + atomic replace, but the snippet still calls SetOutputModelPath with the final compiledModelPath.

**Recommendation:** Show the temporary-path / MoveFileExW flow or the ModelManager::CompileModel helper.

Samples/WindowsML/Shared/cs/ModelManager.cs
adrastogi: Agentic Response: Updated C++, C#, and Python policy-mode validation to follow ONNX Runtime's device ordering, select only the preferred device plus CPU fallback devices, and regroup those selected devices by EP. Cache reuse now also requires matching optimal metadata for the preferred group, so fallback metadata cannot approve a cache for a different accelerator.

Samples/WindowsML/Directory.Packages.props
adrastogi: Agentic Response: Kept 2.3.42 because it is the first managed package exposing the required compatibility APIs, and scoped localpackages to Samples\WindowsML\nuget.config as the optional pre-hydration overlay rather than changing the repository-wide source list. This follows the WASDK team's guidance that the sample-feed hydration pipeline will add the package closure and still permits local validation before hydration.

Samples/WindowsML/cpp/CppConsoleDesktop/README.md
adrastogi: Agentic Response: Updated the snippet to call ModelManager::CompileModel, which is the implementation that compiles to a temporary path, embeds EP context, and atomically replaces the destination with MoveFileExW only after compilation succeeds.

nieubank (approved):

